PR #116, opened 2015-01-05
- fixed switching between btrfs and other filesystems (bsc#910346)

aschnell:


aschnell:
A testcase is included in openQA, see bugreport.

jreidinger:
LGTM

PR #119, opened 2015-02-09
don't set gpt_sync_mbr on Power8

k0da:
Power8 is capable booting off GPT partition table.


aschnell:
LGTM

Needs a port to master of openSUSE/libstorage.

aschnell:
If you urgently need a package/submitrequest in IBS please create one yourself. Otherwise I will do so when I next time work on the package anyway.

k0da:
I need it for next beta

PR #121, opened 2015-02-16
- install into btrfs subvolume (fate#318392)

aschnell:


src/modules/Storage.rb
```diff
@@ -367,13 +368,18 @@ def FinishLibstorage
     end
 
 
+    def DefaultSubvolumeName()
```
ancorgs: I though that the agreement was using Ruby-style names for the new methods, like ```default_subvolume_name```

src/modules/StorageSnapper.rb
```diff
@@ -0,0 +1,151 @@
...
+
+      return tmp.fetch("exit")
+    end
+
+
+    def ConfigureSnapper?()
+
+      part = Storage.GetEntryForMountpoint("/")
+
+      if part.fetch("used_fs", :unknown) != :btrfs
```
ancorgs: I don't think the default ```:unknow``` is needed in this case. This would be more readable, IMO:
```ruby
return false unless part["used_fs"] == :btrfs
```

test/storage_snapper_configure_snapper_test.rb
```diff
@@ -22,13 +22,7 @@
...
     Yast::Storage.stub(:GetEntryForMountpoint).with("/").once.and_return(data)
 
-    Yast::SCR.stub(:Execute).and_return(1)
-    Yast::SCR.should_receive(:Execute).exactly(2).times.and_return(0)
-
-    Yast::SCR.stub(:Write).and_return(1)
-    Yast::SCR.should_receive(:Write).exactly(2).times.and_return(0)
-
-    Yast::StorageUtils.ConfigureSnapper()
+    expect(Yast::StorageSnapper.ConfigureSnapper?).to eq true
```
ancorgs: To be honest, the original tests where not really that awesome :-), but this is clearly reducing the test coverage. Thus making things worse. ```ConfigureSnapper?``` is not the equivalent to the old ```ConfigureSnapper```. We should create tests for the real substitutes (Step1...Step4).

Since we are on it, the tests should be more accurate (not just checking that you call ```:Execute``` without actually checking what you are executing) and more descriptive (the current output looks like "ConfigureSnapper configures snapper. ConfigureSnapper does not configure snapper").

src/modules/StorageSnapper.rb
```diff
@@ -0,0 +1,151 @@
...
+      log.info("configuring snapper for root fs - step 1")
+
+      if ENV["YAST2_STORAGE_SNAPPER_EXTEND"] != "no"
+        # TRANSLATORS: message in progress bar
+        InstExtensionImage.LoadExtension("snapper", _("Retrieving %s extension...") % "snapper")
+      end
+
+      part = Storage.GetEntryForMountpoint("/")
+
+      if BashLogOutput("/usr/lib/snapper/installation-helper --step 1 " <<
```
ancorgs: NP: jfyi, using this would be a little bit more ruby than concatenating:
```ruby
"... step 1 "\
"--device ..."
```

src/modules/StorageSnapper.rb
```diff
@@ -0,0 +1,151 @@
...
+# To contact Novell about this file by physical or electronic mail, you may
+# find current contact information at www.novell.com.
+
+
+require "yast"
+
+
+module Yast
+
+  class StorageSnapperClass < Module
```
ancorgs: Same than before for all the methods in this class, I though that in new code we agreed on using ruby-friendly method names.

test/storage_snapper_configure_snapper_test.rb
aschnell: Testing for the exact execute parameters is a "change detector" and should be avaided, see http://googletesting.blogspot.de/2015/01/testing-on-toilet-change-detector-tests.html. Such tests have negative value and should be removed.

src/modules/Storage.rb
aschnell: OK, renamed although it's not inconsistent to the rest of the class.

src/modules/StorageSnapper.rb
aschnell: I generally dislike those post conditions.

test/storage_snapper_configure_snapper_test.rb
ancorgs: @aschnell Fair enough (although I was actually not talking about checking the whole exact command). But I still can see some things that are not properly tested/described.

For ```ConfigureSnapper?``` we actually have three posibilities (no btrfs, btrfs with snapshots, btrfs without snapshots). Looking at the output of the current test is far from obvious.

For steps 1 and 4, the (un)loading of the extension if needed.

src/modules/StorageSnapper.rb
```diff
@@ -0,0 +1,151 @@
...
+      if userdata.fetch("/", "") != "snapshots"
+        return false
+      end
+
+      return true
+
+    end
+
+
+    def ConfigureSnapperStep1()
```
jreidinger: I know that it is called step 1 in helper, but here I think it would be nicer to use more descriptive and user friendly name, that better explain what is happen there. At least something like `prepare_created_system` as it is called after root filesystem created

src/modules/StorageSnapper.rb
jreidinger: and maybe add there link to installation_helper sources, so it is easier to find what is this helper

test/storage_snapper_configure_snapper_test.rb
aschnell: I have added more tests (that I consider to have low to no value).

src/modules/StorageSnapper.rb
aschnell: All four steps prepare the created system. What they exactly do is up to snapper and I don't see any value in giving descriptive names here. The steps in snapper may change and this would then require also modifications here.

jreidinger: Well, but all steps is called in well defined time like after chroot, after fs created and so not. I think it will be more helpful names then step1..4

aschnell: The translation from those defined times to the snapper steps are done in Storage.rb and storage_finish.rb. For snapper it would be fine if step 2 and 3 are done later. Step 2 might even be a nop if step 3 in snapper changes.

I agree that the names are not perfect but neither are names like snapper_create_config_file_and_create_first_snapshot_and_set_default_subvolume.

jreidinger: well, this way there is four method step1-4 without clear requirements if it can move, if it can be called later or earlier. I think it should be at least documented, what is requirements for such calls, so if someone else want to change it in future he knows what is required and what need to be satisfied when changing.

test/storage_snapper_configure_snapper_test.rb
```diff
@@ -0,0 +1,75 @@
...
+  it "configures snapper" do
+
+    data = {
+      "device" => "/dev/sda1",
+      "mount" => "/",
+      "used_fs" => :btrfs,
+      "userdata" => { "/" => "snapshots" }
+    }
+
+    Yast::Storage.stub(:GetEntryForMountpoint).with("/").once.and_return(data)
```
ancorgs: New syntax
```ruby
allow(Yast::Storage).to receive(:GetEntryForMountpoint).with("/").once.and_return data
```

test/storage_snapper_configure_snapper_test.rb
```diff
@@ -0,0 +1,75 @@
...
+
+ENV["Y2DIR"] = File.expand_path("../../src", __FILE__)
+
+require "yast"
+
+Yast.import "StorageSnapper"
+Yast.import "Storage"
+
+
+describe "StorageSnapper#configure_snapper?" do
```
ancorgs: This syntax for stubbing is deprecated and this still not address the original concern of poor output. Compare this:
```
StorageSnapper#configure_snapper?
  configures snapper
  does not configure snapper
  does not configure snapper
  does not configure snapper

Deprecation Warnings:

Using `stub` from rspec-mocks' old `:should` syntax without explicitly enabling the syntax is deprecated.
Use the new `:expect` syntax or explicitly enable `:should` instead.
Called from /.../storage_snapper_configure_snapper_test.rb:23:in `block (2 levels) in <top (required)>'.
```
to this (just an example, it can be done with less nesting as well):
```
StorageSnapper#configure_snapper?
  when root is btrfs
    when snapshots are enabled
      it returns true
    when snapshots are disabled
      it returns false
  when root is xfs
    when snapshots are enabled
      it returns false
    when snapshots are disabled
      it returns false
```
Anyways, even if we don't agree in the "test output as readable documentation" approach, please fix the deprecation warnings (see below).

test/storage_snapper_configure_snapper_test.rb
```diff
@@ -0,0 +1,75 @@
...
+    data = {
+      "device" => "/dev/sda1",
+      "mount" => "/",
+      "used_fs" => :btrfs,
+      "userdata" => { "/" => "snapshots" }
+    }
+
+    Yast::Storage.stub(:GetEntryForMountpoint).with("/").once.and_return(data)
+
+    expect(Yast::StorageSnapper.configure_snapper?).to eq true
```
ancorgs: NP: this will be sorter using ```subject``` http://betterspecs.org/#subject
That is, at the beginning:
```ruby
 subject { Yast::StorageSnapper.configure_snapper?) }
```
And then just
```ruby
expect(subject).to eq true
```

aschnell: Frankly this rspec syntax is nuts. There's a good way to assign a function to a variable in ruby - but no, rspec does something different. Why not ```subject = lambda { ... }```?

And these pseudo sentences where you either have to use either a space, dot or underscore is just as bad. What's the justification for ```to eq true``` and ``` to be_true```?

src/modules/StorageSnapper.rb
aschnell: Changed, although obeying the coding style is not a requirement.

ancorgs:
LGTM regarding my RSpec concerns. Concerns from @jreidinger about the lack of documentation/meaningful names for methods still stays, I'd say.

src/modules/StorageSnapper.rb
aschnell: Added requirements in snapper.

ancorgs:
Now that the required documentation is in snapper itself and to avoid holding this too much (while @jreidinger is not 100% available), I think we should merge. So LGTM.

PR #122, opened 2015-03-02
- fixed syntax error in last commit

aschnell:


mvidner:
LGTM.
Please comment in https://bugzilla.suse.com/show_bug.cgi?id=916935 about this additional fix.

PR #127, opened 2015-03-16
fix have_ppc_boot to check GPT PReP partitions

k0da:


jreidinger:
update of VERSION file missing. Also it cannot be merged. Otherwise change looks good.

jreidinger:
LGTM, just please make it mergable

PR #130, opened 2015-04-22
fix syntax error (bnc #928046)

gabi2:


mchf:
I miss changelog, otherwise lgtm

PR #131, opened 2015-04-24
Efi cleaning

jreidinger:
proposal how fix can look like. Welcome any comments


libstorage/src/Disk.cc
```diff
@@ -2113,6 +2113,17 @@ int Disk::doRemove( Volume* v )
 	    getStorage()->removeDmMapsTo( getPartDevice(p->OrigNr()) );
 	    ret = v->prepareRemove();
 	    }
+        // before deleting partitions ensure that efi do not contain it
+	if( ret==0 && !p->created() )
+            {
+	    std::ostringstream cmd_line;
+            cmd_line.str("");
```
aschnell: This is useless, see http://www.cplusplus.com/reference/sstream/ostringstream/ostringstream/.

jreidinger: you are right, it is leftover of clearing cmd_line from previous attempt

aschnell:
Changelog is missing.

jreidinger:
ok, I will add it, I firstly like to discuss approach

shundhammer:
The change log looks good for me. As for the functional part of the patch, Arvin already commented on it.

jreidinger:
thanks both of you merging.

PR #133, opened 2015-05-20
- mark partition alignment names for translation (bsc#928949)

aschnell:


storage/src/include/ep-settings.ycp
```diff
@@ -71,8 +77,8 @@
 	    return `item(`id(fs), FileSystems::GetName(fs, "Error"));
 	});
 
-	list<term> partalign_items = maplist(symbol pal, [ `align_optimal, `align_cylinder ], {
-	    return `item(`id(pal), substring( sformat( "%1", pal ), 7 ));
+	list<term> partalign_items = maplist(symbol pal, string label, alignments, {
+	    return `item(`id(pal), label );
```
jreidinger: if it is outside of method, does it handle correctly if you later change locale? ( Like in installation if you choose e.g. spanish, go to partitioning, then back, change to German, if it remain spanish or reevaluated as German? We have such problem in ruby and unless ycp do reevaluation, it affect also ycp as gettext is called only once.

aschnell: AFAIR the translations are reevaluated and the other translations in the file do it exactly the same way.

jreidinger: OK, so LGTM

PR #134, opened 2015-05-22
Add a snapshots_finish client

imobach:
This is a client to create an snapshot after installation/upgrade.


src/lib/storage/snapshots_finish.rb
```diff
@@ -0,0 +1,48 @@
+require "yast2/fs_snapshot"
+require "installation/finish_client"
+
+module Storage
```
jreidinger: There is already storage as libstorage ruby bindings. Are you sure it do not collide?

src/lib/storage/snapshots_finish.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    # Write configuration
+    #
+    # It creates a snapshot when no second stage is required and
+    # Snapper is configured.
+    #
+    # @return [TrueClass,FalseClass] True if snapshot was created;
+    #                                otherwise it returns false.
+    def write
+      if !second_stage_required? && StorageSnapper.configure_snapper?
```
jreidinger: Something is strange, does it work? As this module do not live in Yast namespace, so direct call to StorageSnapper should not work unless you somewhere include Yast namespace inside this class.

src/lib/storage/snapshots_finish.rb
```diff
@@ -0,0 +1,48 @@
...
+module Storage
+  class SnapshotsFinish < ::Installation::FinishClient
+    include Yast::I18n
+
+    def initialize
+      textdomain "storage"
+
+      Yast.import "Mode"
+      Yast.import "StorageSnapper"
+      Yast.include self, "installation/misc.rb"
```
jreidinger: if class do not live in yast namespace, it can be problem, as usually includes expect to live here.

src/lib/storage/snapshots_finish.rb
```diff
@@ -0,0 +1,48 @@
...
+    def title
+      _("Creating root filesystem snapshot...")
+    end
+
+    private
+
+    def create_snapshot(snapshot_type, description)
+      Yast2::FsSnapshot.create_single(description)
+      true
+    rescue Yast2::SnapperNotConfigured, Yast2::PreviousSnapshotNotFound, Yast2::SnapshotCreationFailed
```
jreidinger: SnapperNotConfigured should not happen

src/lib/storage/snapshots_finish.rb
```diff
@@ -0,0 +1,48 @@
...
+      log.error("Filesystem snapshot could not be created.")
```
jreidinger: at least log msg of exception, otherwise it can he hard to guess which one is raised.

src/lib/storage/snapshots_finish.rb
imobach: Yes, I agree, I prefer that class living on the Yast namespace. But I took a look at this and I though it was some kind of agreement: https://github.com/yast/yast-bootloader/blob/master/src/lib/bootloader/finish_client.rb

test/Makefile.am
```diff
@@ -2,9 +2,10 @@
 # Makefile.am for storage/test
 #
 
-TESTS =									\
-	storage_snapper_configure_snapper_test.rb			\
-	storage_get_disk_partition.rb
+TESTS = \
+  storage_snapper_configure_snapper_test.rb \
+  storage_get_disk_partition.rb \
+  storage/snapshots_finish_test.rb
```
jreidinger: Makefile require tabs, here it looks like it is expanded, does it work?

test/storage/snapshots_finish_test.rb
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env rspec
+
+require_relative "../test_helper"
+require_relative "../../src/lib/storage/snapshots_finish"
```
jreidinger: this should not be needed, simple `require "storage/snapshots_finish"` should be enough

test/Makefile.am
imobach: Yes, it works. Just like this file: https://github.com/yast/yast-storage/blob/master/src/Makefile.am

src/lib/storage/snapshots_finish.rb
imobach: Probably they're polluting the same namespace. See below.

src/lib/storage/snapshots_finish.rb
imobach: It's true.

src/lib/storage/snapshots_finish.rb
jreidinger: I have no problem if something do not live in Yast namespace. I just want to point out, that original include code can have problem if not living there.

src/lib/storage/snapshots_finish.rb
```diff
@@ -0,0 +1,48 @@
...
+    # It creates a snapshot when no second stage is required and
+    # Snapper is configured.
+    #
+    # @return [TrueClass,FalseClass] True if snapshot was created;
+    #                                otherwise it returns false.
+    def write
+      if !second_stage_required? && StorageSnapper.configure_snapper?
+        log.info("Creating root filesystem snapshot")
+        action = Mode.update ? "upgrade" : "installation"
+        create_snapshot(:single, "after #{action}")
```
aschnell: During upgrade a post snapshot has to be created.

src/lib/storage/snapshots_finish.rb
```diff
@@ -0,0 +1,48 @@
...
+      end
+    end
+
+    def title
+      _("Creating root filesystem snapshot...")
+    end
+
+    private
+
+    def create_snapshot(snapshot_type, description)
```
jreidinger: snapshot  type is unused variable

src/lib/storage/snapshots_finish.rb
aschnell: StorageSnapper.configure_snapper? does not work during upgrade since it looks at the userdata for the root volume. The required data therein is set by the storage proposal and expert partitioner but so far not by the update code.

jreidinger: this client will be called before unmount of target from /mnt. I hope at this time snapper should be already configured at target location as all packages are already upgraded and everything should be written down. Where exactly it is written to target disk?

src/lib/storage/snapshots_finish.rb
imobach: I've though about implementing that as part of this PBI https://trello.com/c/E8LbHch1/143-5-12-sp1-m-317900-create-snapshot-before-starting-upgrade. I guess I need the number of the `pre` snapshot to pair with it.

src/lib/storage/snapshots_finish.rb
imobach: Yeah, dead code.

src/lib/storage/snapshots_finish.rb
imobach: Should FsSnapshot.configure? do the trick? https://github.com/yast/yast-yast2/blob/master/library/system/src/lib/yast2/fs_snapshot.rb#L72

imobach:
The `SnapshotsFinish` class have been moved to `Yast::YStorage`. Why?

1. I don't like polluting the global namespace (I missunderstood [this definition](yast/yast-bootloader/blob/master/src/lib/bootloader/finish_client.rb) and I thought that there was some kind of agreement about that).
2. @jreidinger showed some concerns about using `Yast.include` outside of `Yast` namespace. It seems to work but it's better to make sure we don't break anything.
3. I think that `Yast::Storage` is the right place for `SnapshotsFinish` to live but, as @jreidinger pointed out, there is a constant called `Yast::Storage`.

So I thought that adding a new namespace won't hurt.

jreidinger:
Looks, good, I am just creating counter part in pre update, that create entry in store.

jreidinger:
LGTM

PR #135, opened 2015-05-28
Removed SnapshotsFinish client

imobach:
* Moved to yast-installation.
* As the package was never released, I'm reverting changes in packaging.


jreidinger:
LGTM

PR #136, opened 2015-06-09
Fixed bsc#922308: No snapper config for LVM storage proposal

shundhammer:
- Added logging to the PostProcessor so it's much easier to check in the logs if making snapshots was even considered there

- Added the missing PostProcessor call to that code path in question

### DIFF
--- a/package/yast2-storage.changes
+++ b/package/yast2-storage.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Jun  9 11:41:38 CEST 2015 - shundhammer@suse.de
+
+- No snapper config for LVM storage proposal (bsc#922308)
+- version 3.1.56
+
+-------------------------------------------------------------------
 Wed May 06 16:34:04 CEST 2015 - aschnell@suse.de
 
 - require rubygems of default ruby version (bsc#929899)
--- a/src/modules/StorageProposal.rb
+++ b/src/modules/StorageProposal.rb
@@ -4440,7 +4440,10 @@ module Yast
 
       end
 
+
       def modify(partitions)
+
+        Builtins.y2milestone( "Post-processing partitions..." )
 
         partitions.each do |volume|
 
@@ -4475,6 +4478,9 @@ module Yast
               size_limit_k = 1024 * opts["root_base"]
               if volume["size_k"] >= size_limit_k
                 volume["userdata"] = { "/" => "snapshots" }
+                Builtins.y2milestone( "Adding userdata '/' => 'snapshots'" )
+              else
+                Builtins.y2milestone( "Below size limit - NOT adding userdata '/' => 'snapshots'" )
               end
             end
           end
@@ -6095,7 +6101,14 @@ module Yast
         )
         Builtins.y2milestone("get_proposal_vm sol:%1", disk)
       end
-      Builtins.y2milestone("get_proposal_vm ret:%1", ret)
+
+      post_processor = PostProcessor.new()
+      ret["target"] = post_processor.process_target(ret["target"])
+
+      Builtins.y2milestone(
+        "get_proposal_vm ret[ok]: %1",
+        Ops.get_boolean(ret, "ok", false)
+      )
       deep_copy(ret)
     end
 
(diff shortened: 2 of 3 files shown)


package/yast2-storage.changes
```diff
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Jun  9 11:41:38 CEST 2015 - shundhammer@suse.de
+
+- No snapper config for LVM storage proposal (bsc#922308)
```
lslezak: The change log entry is a bit unclear, was no snapper config fixed or do we want to have no snapper config for LVM? (It's clear from the bug description, but here the wording is a bit misleading, at least for me...)

src/modules/StorageProposal.rb
```diff
@@ -4440,8 +4440,11 @@ def analyse(partitions)
 
       end
 
+
       def modify(partitions)
 
+        Builtins.y2milestone( "Post-processing partitions..." )
```
ancorgs: Please, use log.info (new way) instead of y2milestone (deprecated).

src/modules/StorageProposal.rb
```diff
@@ -4475,6 +4478,9 @@ def modify(partitions)
               size_limit_k = 1024 * opts["root_base"]
               if volume["size_k"] >= size_limit_k
                 volume["userdata"] = { "/" => "snapshots" }
+                Builtins.y2milestone( "Adding userdata '/' => 'snapshots'" )
```
ancorgs: Same that above. In addition, spaces after the "(" are against the Ruby coding standard.

src/modules/StorageProposal.rb
```diff
@@ -6095,7 +6101,14 @@ def get_proposal_vm(target, key, disk)
...
       end
-      Builtins.y2milestone("get_proposal_vm ret:%1", ret)
+
+      post_processor = PostProcessor.new()
+      ret["target"] = post_processor.process_target(ret["target"])
+
+      Builtins.y2milestone(
+        "get_proposal_vm ret[ok]: %1",
+        Ops.get_boolean(ret, "ok", false)
+      )
```
ancorgs:
```ruby
log.info "get_proposal_vm ret[ok]: #{ret['ok']}"
```

src/modules/StorageProposal.rb
lslezak: For the new code we should prefer to use the more Ruby way for logging, see http://www.rubydoc.info/github/yast/yast-ruby-bindings/master/Yast/Logger.

src/modules/StorageProposal.rb
```diff
@@ -6095,7 +6101,14 @@ def get_proposal_vm(target, key, disk)
         )
         Builtins.y2milestone("get_proposal_vm sol:%1", disk)
       end
-      Builtins.y2milestone("get_proposal_vm ret:%1", ret)
+
+      post_processor = PostProcessor.new()
```
lslezak: NP: style: simple `PostProcessor.new` is better.

src/modules/StorageProposal.rb
```diff
@@ -6095,7 +6101,14 @@ def get_proposal_vm(target, key, disk)
...
+      ret["target"] = post_processor.process_target(ret["target"])
+
+      Builtins.y2milestone(
+        "get_proposal_vm ret[ok]: %1",
+        Ops.get_boolean(ret, "ok", false)
```
lslezak: Avoid `Ops.*` methods in the new code if possible, they are just wrappers for the old YCP builtins/operators. Use the native Ruby code instead (simple `ret["ok"]`).

shundhammer: This is copied verbatim from the other places where this is used. Since there is no reasonable way to test this isolated, I tried to be minimal-invasive with those changes.

package/yast2-storage.changes
shundhammer: That's the bug title.

aschnell:
The code style remark are nice-to-have (http://yastgithubio.readthedocs.org/en/latest/code-review/).

LGTM

PR #137, opened 2015-06-18
Verbose message when bios_grub partition is missing (bnc#917763)

ancorgs:


src/include/partitioning/custom_part_check_generated.rb
```diff
@@ -336,9 +336,12 @@ def check_created_partition_table(targetMap, installation)
       if (!boot_found && installation && 
           !Partitions.EfiBoot && rootdlabel=="gpt") || show_all_popups
         message = _(
-          "Warning: There is no partition of type bios_grub present.\n" +
-          "To boot from a GPT disk using grub2 such a partition is needed.\n" +
-          "\n" +
+          "Warning: There is no partition of type bios_grub present.\n" \
+          "When installing Grub2 to the MBR of a GPT formatted disc it is recommended\n" \
+          "(if Btrfs is used, then it is required) to reserve a small (around 1MB)\n" \
+          "unformatted partition of this type.\n" \
```
mchf: This sentence is killing me - spread over three lines and many braces in it ;-)

Btw this seems to be pretty long for a popup. It looks more like a help text.

ancorgs: Well,the sentence is taken from the bug report. And the report is precisely about having a more informative pop-up. I'll try to make the sentence shorter and more readable without losing the information the reporter wanted to have there.

ancorgs: @mchf What about?
```ruby
"Warning: There is no partition of type bios_grub present.\n" \
"Such a partition is recommended (required with Btrfs) when Grub2 is\n" \
"installed in the MBR of a GPT disk. It must be unformatted and\n" \
"approximately 1MB large." \
"\n" \
"Really use this setup?\n"

File: package/yast2-storage.changes [hunk 2]
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Jun 18 09:47:16 UTC 2015 - ancor@suse.com
+
+- Improved the message displayed when a bios_grup partition is
```
kobliha: Or maybe `bios_grub`?

src/include/partitioning/custom_part_check_generated.rb
mchf: yep, much better. LGTM

package/yast2-storage.changes
ancorgs: Yep. Good catch

src/include/partitioning/custom_part_check_generated.rb
```diff
@@ -336,9 +336,12 @@ def check_created_partition_table(targetMap, installation)
       if (!boot_found && installation && 
           !Partitions.EfiBoot && rootdlabel=="gpt") || show_all_popups
         message = _(
-          "Warning: There is no partition of type bios_grub present.\n" +
-          "To boot from a GPT disk using grub2 such a partition is needed.\n" +
-          "\n" +
+          "Warning: There is no partition of type bios_grub present.\n" \
+          "When installing Grub2 to the MBR of a GPT formatted disc it is recommended\n" \
```
kobliha: when ... formatted disc>>,<< it is recommended

src/include/partitioning/custom_part_check_generated.rb
```diff
@@ -336,9 +336,12 @@ def check_created_partition_table(targetMap, installation)
...
+          "(if Btrfs is used, then it is required) to reserve a small (around 1MB)\n" \
```
kobliha: (required in case if btrfs partition)

kobliha:
Nice to have such a message but I'm quite lost as well :)
Can't we detect whether, e.g., btrfs is used and then just show different message?
Can we check whether GPT is used, ... can we...?

src/include/partitioning/custom_part_check_generated.rb
aschnell: We use "disk", not "disc".

ancorgs: @aschnell, @kobliha You are reviewing the original text that was already agreed to be... let's say... suboptimal. Please review the one proposed in https://github.com/yast/yast-storage/pull/137#discussion_r32722471

ancorgs:
@kobliha, we already know that GPT is used. Nevertheless I find the "of a GPT disk" part informative.

It's more or less the same with btrfs, we could check for the filesystem and write (recommended|required) depending on the case, but I (as a user) would appreciate that the dialog explains me the difference, no matter what I selected.

mchf:
LGTM

PR #138, opened 2015-06-18
Default to no copy on write for MariaDB on btrfs (fate#316718)

shundhammer:


aschnell:
LGTM, except travis failure

PR #139, opened 2015-06-26
- added Storage::SetUserdata() (for bsc#935858)

aschnell:


jreidinger:
in general LGTM, I just do not see direct reason to add such method...usage from autoyast?

aschnell:
Yes, to be used in AutoYaST.

src/modules/Storage.rb
```diff
@@ -2706,8 +2706,19 @@ def SetCrypt(device, crpt, format)
...
 
 
-    def ChangeDescText(dev, txt)
-      ret = @sint.changeDescText(dev, txt)
+    def ChangeDescText(device, text)
+      @sint.changeDescText(device, text)
+    end
+
+
+    def SetUserdata(device, userdata)
```
kobliha: Can we have a description for this? @param @return ...

PR #140, opened 2015-06-30
- disallow using unpartitioned DASDs for LVM (fate #317934)

aschnell:


src/include/partitioning/lvm_pv_lib.rb
```diff
@@ -128,9 +128,11 @@ def get_possible_pvs(targetMap)
           )
         end
         if Builtins.size(Ops.get_list(devmap, "partitions", [])) == 0 &&
-            Storage.IsPartType(Ops.get_symbol(devmap, "type", :CT_UNKNOWN)) &&
-            (!Storage.IsUsedBy(devmap) ||
-              Ops.get_symbol(devmap, "used_by_type", :UB_NONE) == :UB_LVM)
+           Storage.IsPartType(Ops.get_symbol(devmap, "type", :CT_UNKNOWN)) &&
+           # unpartitioned DASDs cannot be used for LVM (fate #317934)
+           !String.StartsWith(dev, "/dev/dasd") &&
```
jreidinger: so this info is not in devmap and only we can get it by checking prefix of def....maybe more ruby `dev.start_with?("/dev/dasd")` is better?

jreidinger:
just idea for improvement, up to you. LGTM

PR #141, opened 2015-07-14
Code 11 sp4 bug 936187

schubi2:
loading the correct multpath module in storage module


aschnell:
Looks per se fine, although the code should be obsolete since libstorage loads dm-multipath when activating multipath.

aschnell:
Basically a backport from master. LGTM

PR #143, opened 2015-07-15
Allow /boot on RAID1 for PowerPC

k0da:
PowerPC bootloaders are able to read raid1 partition

Fixes bsc#938170

### DIFF
--- a/package/yast2-storage.changes
+++ b/package/yast2-storage.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Wed Jul 17 13:31:34 CEST 2015 - dvaleev@suse.com
+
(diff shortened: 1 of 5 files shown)


package/yast2-storage.changes
```diff
@@ -10,7 +16,6 @@ Tue Jul 14 15:27:09 CEST 2015 - schubi@suse.de
 
 - Loading the correct multipath module (dm-multipath)
   (bnc#937942)
-- 3.1.61
```
jreidinger: this looks wrong, Please do not remove any entry from old change log

jreidinger:
@k0da please just return back changelog line, then we can merge it.
Thanks

jreidinger:
thanks for fix

PR #144, opened 2015-07-16
- do not try to load snapper extension since it does not exist anymore (see bsc#937243)

aschnell:


ancorgs:
LGTM

mvidner:
LGTM.
BTW there seem to be more places that load snapper, this is one: https://github.com/yast/yast-yast2/blob/0eeb37a3cd96ee7cf4f54cc8daff0ead851d0707/library/system/src/lib/yast2/fs_snapshot.rb

aschnell:
Yes, there are more places, see https://bugzilla.suse.com/show_bug.cgi?id=938377.

PR #145, opened 2015-07-23
Added NoCoW subvolume for /var/lib/libvirt/images

shundhammer:
https://fate.suse.com/319299


jsrain:
LGTM except the failed check

PR #146, opened 2015-08-07
Fix special /boot handling on SLE12

imobach:
It fixes https://bugzilla.suse.com/show_bug.cgi?id=940374. The real problem is that, although the installer warns about converting automatically /boot partition to PReP, it does not nothing about that (on expert partitioner).


ancorgs:
LGTM

PR #147, opened 2015-08-07
Fix special /boot handling on SLE12 SP1

imobach:
The same fix that PR https://github.com/yast/yast-storage/pull/146 but for SLE12 SP1. Manually tested on ppc64le and x86_64.


ancorgs:
LGTM

PR #148, opened 2015-08-10
don't create btrfs subvolumes outside the root fs (bsc#940797)

wfeldt:
Although yast already warns about such a setup it would be nice not to propose
such a setting in the first place.
The code is inspired by check_created_partition_table().

### DIFF
--- a/src/modules/Storage.rb
+++ b/src/modules/Storage.rb
@@ -4978,6 +4978,28 @@ module Yast
     end
 
 
+    def SubvolShadowed?(subvol)
+      shadowed = false
+      tmp = "/" + subvol
+
+      GetTargetMap().each do |dev, disk|
+        parts = disk.fetch("partitions", [])
+        parts.each do |part|
+          if part.has_key?("mount")
+            mp = part["mount"]
+            if (tmp == mp) || (tmp.start_with?(mp) && tmp[mp.size] == "/")
+              shadowed = true
+            end
+          end
+        end
+      end
+
+      Builtins.y2milestone("SubvolShadowed?( %1 ): %2", subvol, shadowed)
+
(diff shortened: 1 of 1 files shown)


wfeldt:
The is a preliminary patch. Although it fixes the issue, I've no idea if that conforms to the expected style or if such a fix is even wanted.

aschnell:
AFAIS the new behaviour can also surprise users and cause trouble: E.g. you add /home partition, then / (and no /home subvolume) and finally you decide to remove the /home partition again. Now /home will be reverted during a rollback since no subvolume exists for it.

src/modules/Storage.rb
```diff
@@ -4978,6 +4978,28 @@ def DetectHomeFs(p)
     end
 
 
+    def SubvolShadowed?(subvol)
```
kobliha: Please, add a function description: What it does and what it actually returns.

wfeldt:
Well, when you configure a separate /home you don't get a /home subvolume. So that would be in
favor of my argument for no /boot subvolumes. Or not?

aschnell:
I don't think the comparison with /home help here since it's a perfect fine setup to have /home on its one btrfs filesystem and still have subvolumes for all users, e.g. /home/snwint.

wfeldt:
Well, sure. But then they are subvolumes of the /home file system.. Here we are talking about btrfs subvolumes of / mounted over an ext4 /boot.

jreidinger:
@aschnell @wfeldt what is status here?

aschnell:
FMPOV to merge the pull request it needs more discussion. Since the bug is already closed I don't mind if the pull request is closed without merging.

PR #149, opened 2015-08-19
Btrfs on lvm

kobliha:
Doesn't pass the test case in build with `Can't import StorageCallbacks`, but it works on the running system.


kobliha:
Hmm, according to Arvin, this must not go to SLE 12, so only 42 and SLE 12 SP1

PR #150, opened 2015-08-25
Install storage related packages only on demand

shundhammer:
Install storage related packages only on demand
(bsc#937040, bsc#907331, bsc#915097) 

https://trello.com/c/NQRRvni6/106-8-sles12-sp1-install-storage-related-packages-only-on-demand

### DIFF
--- a/package/yast2-storage.changes
+++ b/package/yast2-storage.changes
@@ -1,8 +1,7 @@
 -------------------------------------------------------------------
 Tue Aug 25 16:21:55 CEST 2015 - shundhammer@suse.de
 
-- Install storage related packages only on demand
-  (bsc#937040, bsc#907331, bsc#915097) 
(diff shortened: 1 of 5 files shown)


package/yast2-storage.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Aug 25 16:21:55 CEST 2015 - shundhammer@suse.de
+
+- Install storage related packages only on demand
+  (bsc#937040, bsc#907331, bsc#915097)
```
aschnell: bsc #907331 and #915097 are not fixed by this.

aschnell:
Since we also want to fix bug #907331 we need the list of filesystem packages here. And since installation from supplements is still not possible in ruby (see fate #305913) we cannot improve it now.

aschnell:
LGTM

PR #151, opened 2015-08-26
Test case for new Storage.SetUserData function

kobliha:
@shundhammer This is just a test case from already LGTMed PR #149 
The code is already in SP1 / master


shundhammer:
LGTM

PR #153, opened 2015-09-08
Fix boot handling

imobach:
This PR should fix https://bugzilla.suse.com/show_bug.cgi?id=944431

This is a WIP (as further testing is needed), but I'd like to have some feedback from @aschnell or @shundhammer. So please, don't hesitate to comment.

I don't know if something needs to be done for this case: https://github.com/yast/yast-storage/blob/master/src/modules/Storage.rb#L6177

Finally, I've manually tested on x86_64, but that's not enough. More testing to come!

### DIFF
--- a/src/modules/Storage.rb
+++ b/src/modules/Storage.rb
@@ -6206,7 +6206,6 @@ module Yast
       deep_copy(tg)
     end
 
-
     def PerformLosetup(loop, format)
       crypt_ok = false
       pwd = Ops.get_string(loop.value, "passwd", "")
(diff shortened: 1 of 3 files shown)


src/modules/Storage.rb
```diff
@@ -7126,6 +7125,42 @@ def GetDmraidToMdadm()
...
+          # fsid
+          if old_part["fsid"] != part["fsid"]
+            SetPartitionId(part["device"], part["fsid"])
+          end
+
+          # mount point
+          if old_part["mount"] != part["mount"]
+            SetPartitionMount(part["device"], part["mount"])
+          end
+        end
```
imobach: As long as only /boot partition is supposed to exist, I guess we could break the loop when we found it.

Rakefile
```diff
@@ -3,4 +3,5 @@ require "yast/rake"
 Yast::Tasks.configuration do |conf|
   #lets ignore license check for now
   conf.skip_license_check << /.*/
+  conf.obs_target = "SLE_12"
```
schubi2: That's your favorite. Right ? :-)

imobach: Argh! Thanks for noticing. I'll put a Git hook to avoid this! :)

aschnell:
It looks technical OK but I'm unsure this is overall the right way to go.

Instead it should be possible to create a PReP partition in the expert partitioner (cannot test right now whether that works) and then add checks to src/include/partitioning/custom_part_check_generated.rb if a PReP partition is missing.

imobach:
@aschnell Then your approach will be to warn the user about the missing partition and let her to set it up. It's fine.

However I wonder if [this message](https://github.com/yast/yast-storage/blob/master/src/include/partitioning/custom_part_check_generated.rb#L364) will be still valid after we disable the SpecialBootHandling stuff (maybe the conversion is done in another place).

BTW, I saw that OpenQA guys have fixed the partitioning RAID test so the PReP partition is created by the user.

aschnell:
That message will then not be valid and apparently also was not valid in the past (maybe it was valid may years ago). When was the openQA test created? Did it work in the past?

imobach:
I'm not sure when the test was created, but I saw a failure from around [2 months ago](https://openqa.suse.de/tests/104827).

imobach:
@aschnell @kobliha So that's the plan (if you agree):

1. Revert the offending commit.
2. Write a unit test for [check_created_partition_table](https://github.com/yast/yast-storage/blob/master/src/include/partitioning/custom_part_check_generated.rb#L59) method to avoid breaking more stuff.
3. Fix the /boot detection during checks. When a `/boot` [partition is found](https://github.com/yast/yast-storage/blob/master/src/include/partitioning/custom_part_check_generated.rb#L160), the PReP/CHRP check will not run.

aschnell:
Item 3 looks wrong to me. AFAIK PPC always needs a PReP partition regardless whether there is a /boot partition. So I would simply warn the user if there is no PReP partition. But better ask the PPC team.

imobach:
@aschnell Sorry, but I expressed badly: what I wanted to say is that the check does not run properly for the time being. And we need to fix that.

aschnell:
So can we close this pull request since you will make a new one with the improved checks?

imobach:
@aschnell Yes, I expect to submit a new pull request in a while. Closing this one. Thanks.

PR #154, opened 2015-09-11
Fix LVM on expert partitioner

imobach:
* Reverts 29b5c41, which broke LVM stuff (and who knows what other stuff).
* Find a different solution for bug ([bsc#940374](https://bugzilla.suse.com/show_bug.cgi?id=940374) as @aschnell suggested).
* Fixes the missing PReP/CHRP message. The old message stated that `/boot` partitions would be converted to PReP/CHRP automatically. That's not true.

### DIFF
--- a/src/include/partitioning/custom_part_check_generated.rb
+++ b/src/include/partitioning/custom_part_check_generated.rb
@@ -70,6 +70,7 @@ module Yast
       partition_mounted_but_not_formated = false
       swap_found = false
       boot_found = false
+      boot_mount_point = ""
       root_found = false
       gpt_boot_ia64 = false
       boot_end = 0
@@ -158,6 +159,7 @@ module Yast
               gpt_boot_ia64 = true
             end
             boot_found = true
+            boot_mount_point = mountpoint
             if Ops.get_symbol(diskinfo, "type", :CT_UNKNOWN) == :CT_DISK
               boot_end = Region.End(Ops.get_list(part, "region", []))
             else
@@ -174,6 +176,7 @@ module Yast
             if Partitions.PrepBoot &&
                 (fsid == Partitions.FsidBoot(dlabel) || fsid == 6)
               boot_found = true
+              boot_mount_point = mountpoint
               boot_end = Region.End(Ops.get_list(part, "region", []))
               boot_fs = Ops.get_symbol(part, "used_fs", :unknown)
               boot_size_k = Ops.get_integer(part, "size_k", 0)
@@ -181,11 +184,13 @@ module Yast
             elsif Arch.board_mac &&
                 Ops.get_symbol(part, "used_fs", :unknown) == :hfs
               boot_found = true
+              boot_mount_point = mountpoint
               boot_end = Region.End(Ops.get_list(part, "region", []))
               boot_fs = Ops.get_symbol(part, "used_fs", :unknown)
               boot_size_k = Ops.get_integer(part, "size_k", 0)
             elsif !Partitions.EfiBoot() && fsid == Partitions.fsid_bios_grub
               boot_found = true
+              boot_mount_point = mountpoint
               boot_end = Region.End(Ops.get_list(part, "region", []))
               boot_fs = :none
               boot_size_k = Ops.get_integer(part, "size_k", 0)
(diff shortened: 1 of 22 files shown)


aschnell:
Surprised to see YCP.

LGTM

imobach:
@aschnell I was considering about using JSON, but YCP was pretty straigthforward. And it also support symbols.

Thanks for reviewing.

src/include/partitioning/custom_part_check_generated.rb
```diff
@@ -294,6 +299,20 @@ def check_created_partition_table(targetMap, installation)
         ok = false if !Popup.YesNo(message)
       end
 
+      # A PReP/CHRP partition is not supposed to be mounted. So if we find any
+      # other /boot partition, we should warn the user.
+      if boot_found && Partitions.PrepBoot && !boot_mount_point.empty? && !Arch.board_iseries && installation || show_all_popups
```
kobliha: @imobach Although this probably works, it's quite a tricky condition ;) BTW why do we have !Arch.board_iseries there? It's partly documented, and this is missing. Shouldn't we rather use Mode.installation?

src/include/partitioning/custom_part_check_generated.rb
```diff
@@ -360,9 +379,8 @@ def check_created_partition_table(targetMap, installation)
             _(
               "Warning: There is no partition mounted as /boot.\n" +
                 "To boot from your hard disk, a small /boot partition\n" +
-                "(approx. %1) is required.  Consider creating one.\n" +
-                "Partitions assigned to /boot will automatically be changed to\n" +
-                "type 0x41 PReP/CHRP.\n" +
```
kobliha: @imobach Has the behavior changed? It stated `Partitions assigned to /boot will automatically be changed...` and now it doesn't do so anymore? If so, we might need to tell the docu team.

test/include/partitioning_custom_part_check_generated_include_test.rb
```diff
@@ -0,0 +1,977 @@
+#!/usr/bin/env rspec
```
kobliha: @imobach Is the test called during build? I don't think so, as it's not mentioned in Makefile...

src/include/partitioning/custom_part_check_generated.rb
imobach: Yep, it's really tricky, but iSeries seems to be an special case and no PReP partition is needed. Take a look, for example, to this condition: https://github.com/imobach/yast-storage/blob/7fbee7ff1355ffd20d409c018a1b525bdb22469a/src/include/partitioning/custom_part_check_generated.rb#L374

On the other hand, installation is a parameter to this function.

src/include/partitioning/custom_part_check_generated.rb
imobach: Quoting @aschnell: *That message will then not be valid and apparently also was not valid in the past (maybe it was valid may years ago).*

Who's the right person to talk about this change?

kobliha: Hmm, I'd start with @jsrain then `docu` team. It's probably not documented at all.

test/include/partitioning_custom_part_check_generated_include_test.rb
imobach: Ups, I didn't notice that we have a Makefile.am there. It's fixed [in this PR](https://github.com/yast/yast-storage/pull/156).

src/include/partitioning/custom_part_check_generated.rb
kobliha: @imobach Then I'd probably put it out in a separate function `boot_partition_needed?` - I know it's now a one-liner, but in this one-liner, you could document it and reuse it from other places. That's DRY and makes the code readable. It's just a hint...

imobach: Yes, you're right, it would be nice. I tried to be as conservative as possible but I think it worth making the change. I'll submit a PR.

kobliha: Thank you, sir :)

kobliha:
BTW, it's great to have so extensive test!

imobach:
I agree :) (specially in such a critical module).

PR #155, opened 2015-09-11
Propose /boot for encrypted root on PowerPC

k0da:


aschnell:
As I already wrote in bugzilla this is a hack. For example if the proposal reuses partitions the ordering of the two /boot partitions generated in the proposal may change and then the sizes of /boot and PReP are also swapped. And a /boot partition with 7MiB to too small to even hold a filesystem.

k0da:
aren't boot and boot2 different, boot will always be passed as first hence converted.

aschnell:
You likely also missed the code in get_proposal_vm().

aschnell:
It's not obvious that boot will always be passed as first and is always the one that has to be converted to PReP. Can you show that?

Even if, it's still a hack and we are required to improve code quality and not add one more hack. Things have changed in the last years.

k0da:
We could improve SpecialBoothandling by looking at partition size, if it less than 2 cyls then convert it

aschnell:
That is also a hack. The proper fix is to make the proposal code output the correct partitions from the beginning (as already stated in bugzilla).

testsuite/tests/empty-ppc64le4.rb
```diff
@@ -0,0 +1,36 @@
+# encoding: utf-8
+
+# testedfiles: helper.rb
+
+module Yast
```
kobliha: I'm missing a description of this test. Please add at least a comment what it tests and how. And what is the expected output.

kobliha:
I agree with Arvin, that hacks must not be added to the code anymore. I'm expecting a cleanup rather than adding something that will take five times more to fix later. It might help to have a chat with Arvin/HuHa at first, before changing the code. It might be faster.

testsuite/tests/empty-ppc64le4.rb
k0da: I don't see any other tests are doing this.

k0da:
@kobliha do you want refactor at RC1?

testsuite/tests/empty-ppc64le4.rb
kobliha: Yes, you are right, I would like to see it for all the tests from now on. Everything is for the first time once...

kobliha:
@k0da No, what I want is a working code that does not add more hacks. I know that you want to help us to "make it work", but please, consult with @aschnell and @shundhammer how to make it properly. As I've already written in Bugzilla, we need a specification for location and appearance of the PReP partition before implementing anything. We can't be sure that it will work otherwise.

src/modules/StorageProposal.rb
```diff
@@ -5624,11 +5624,16 @@ def get_inst_prop_vm(target, key)
       Ops.set(ret, "target", target)
 
       boot2 = {}
-      if GetProposalEncrypt() && Partitions.EfiBoot
+      if GetProposalEncrypt() && (Partitions.EfiBoot||Partitions.PrepBoot)
```
kobliha: BTW, and this is pure nitpicking, Ruby style requires spaces around `||`

teclator:
@k0da do you plan to do something with this PR, it is currently outdated and the related bug is already closed so probably we could also close it.

jreidinger:
storage is replaced by storage-ng so closing.

PR #156, opened 2015-09-14
Add PartitioningCustomPartCheckGeneratedInclude tests to Makefile

imobach:
Add PartitioningCustomPartCheckGeneratedInclude to Makefile.am so they're executed during package build.


imobach:
@shundhammer I think yes. In RSpec is common to make the name of the file match the name of the class you're testing. And the class is called: `PartitioningCustomPartCheckGeneratedInclude`.

Ideally the class should be called like `Partitioning::CustomPartCheckGenerated` or something like that, but that's not the case at this point.

aschnell:
So the long name is a fallout of YCP to Ruby conversation :(

shundhammer:
Sigh... so again we have to do something that will hurt us for a long time to make some tool happy. :-(

shundhammer:
If we have to do it that way, this looks okay to me then (not "good"...).

imobach:
@shundhammer I'd say that it's the other way around: we have not-so-nice code (a class named like that it's really weird) and we have some side effects like that. As @aschnell it's something that comes from the conversion. But hopefully it will be fixed in the next generation of the storage layer :)

PR #157, opened 2015-09-15
Move check about PReP boot needed to Partitions module

imobach:
This is a minor refactoring [proposed by @kobliha](https://github.com/yast/yast-storage/pull/154#discussion_r39385283).

### DIFF
--- a/src/include/partitioning/custom_part_check_generated.rb
+++ b/src/include/partitioning/custom_part_check_generated.rb
@@ -301,7 +301,7 @@ module Yast
 
       # A PReP/CHRP partition is not supposed to be mounted. So if we find any
       # other /boot partition, we should warn the user.
-      if boot_found && Partitions.PrepBoot && !boot_mount_point.empty? && !Arch.board_iseries && installation || show_all_popups
+      if boot_found && Partitions.prep_boot_needed? && !boot_mount_point.empty? && installation || show_all_popups
         message = _(
           "Warning:\n" \
           "Your system needs a boot partition with type 0x41 PReP/CHRP.\n" \
@@ -369,9 +369,7 @@ module Yast
       end
 
       if !boot_found && installation || show_all_popups
-        # iSeries does not really need a boot partition
-        # a bootable binary will be written to a kernel slot in /proc
-        if Partitions.PrepBoot && !Arch.board_iseries && !diskless || show_all_popups
+        if Partitions.prep_boot_needed? && !diskless || show_all_popups
           # popup text
           # If the user chooses 'no' here, the system will not be able to
           # boot from the hard drive!
--- a/src/modules/Partitions.rb
+++ b/src/modules/Partitions.rb
@@ -997,6 +997,17 @@ module Yast
       text
     end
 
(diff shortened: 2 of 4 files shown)


aschnell:
Having NeedPrep and prep_boot_needed? is confusing. When should which function be used? Clarification from PPC team seems necessary.

imobach:
@aschnell I guess you meant `PrepBoot` (as I don't see any `NeedPrep` method). Or maybe `NeedBoot`?

I guess that `PrepBoot` determines whether the system should boot from a *PRep/CHRP* partition. And the intention behind `prep_boot_needed?` is to check whether is mandatory to have a *PReP/CHRP* boot partition (when as far as I know, it's not mandatory for iSeries).

@k0da: could you help us to clarify this, please?

kobliha:
@imobach That's the problem of undocumented API, we have to guess. Or could you find the docu? PLS, when you find out, add the API docu to at least that method :) thanks in advance

k0da:
I wouldn't bother with iSeries. There is no dedicated hardware for iSeries, all runs on Power

imobach:
@k0da So we're proposing handling the iSeries in the same way as we handle any other PPC, isn't it (so PReP partition will be mandatory)?

kobliha:
The code was there for some reason. That's why we have comments and commit messages and Bugzilla and FATE. Please keep the same functionality as before unless we have a reason to remove it. We are JUST before RC1 and we can't afford losing functionality now.

kobliha:
Although there might be no dedicated hardware, Yast code contains this at least 13-times, see https://github.com/search?utf8=%E2%9C%93&q=board_iseries+user%3Ayast&type=Code&ref=searchresults

imobach:
@kobliha Yes, it makes sense. So back to the original question :). These are my guesses:

* PrepBoot determines whether the type of the boot partition should be 0x41 PReP/CHRP.
* For iSeries, this boot partition is not mandatory (as [stated here](https://github.com/yast/yast-storage/blob/master/src/include/partitioning/custom_part_check_generated.rb#L372)).

@k0da Is that correct? Thanks!

aschnell:
If we don't have to support iSeries anymore please make a feature request to drop the support. Otherwise more and more obsolete code piles up and causes work as can be seen here.

k0da:
@imobach sorry, I don't know how iSeries boots

imobach:
@aschnell @kobliha Taking a look to [iSeries: Linux in a guest partition](http://publib.boulder.ibm.com/iseries/v5r2/ic2924/info/rzalm/rzalm.pdf), the PReP partition is not mandatory for iSeries. You have different options to boot the system.

So for me: `PrepBoot` determines whether the system should boot from a PRep/CHRP partition and `prep_boot_needed?` checks whether is mandatory to have a PReP/CHRP boot partition.

kobliha:
LGTM (from non-storage-expert POV)

aschnell:
LGTM, although clarification from PPC team is missing.

imobach:
Thanks to all!

PR #159, opened 2015-09-22
Revert commit 29b5c41 which broke LVM partitioning

imobach:


imobach:
@aschnell I think it's the minimal fix for this bug.

aschnell:
I would mention in the changelog that it reverts the previous change.

Otherwise LGTM.

imobach:
@aschnell Ok, I've added a note in the changes file.

PR #164, opened 2015-10-29
Added /var/lib/mysql subvolume with NoCoW for older package versions of MariaDB/MySQL

shundhammer:


src/modules/Storage.rb
```diff
@@ -5021,6 +5021,7 @@ def AddSubvolRoot(part)
         "var/lib/libvirt/images",
         "var/lib/mailman",
         "var/lib/mariadb",
+        "var/lib/mysql",
```
jreidinger: so both are valid now?

jreidinger:
LGTM

src/modules/Storage.rb
shundhammer: Right. See also explanations in the bug.

sysrich: var/lib/mysql is used by openSUSE and SLE's mariadb packages, so the addition is really required

var/lib/mariadb isn't used by any package I can find, so it's kind of useless

But, someday that might change (there's been talk by debian for example to start using /var/lib/mariadb as newer versions are not compatable with mysql) so keeping it around doesn't cost anything

shundhammer: /var/lib/mariadb was explicitly demanded in that underlying feature request, so who are we to simply remove it... ;-)

PR #170, opened 2015-11-20
Fix boo#955103

alanbortu:
Resolves issue regarding YaST2 partitioning module allowing you to enter an empty subvolume name.


src/include/partitioning/custom_part_dialogs.rb
```diff
@@ -1674,8 +1674,7 @@ def SubvolHandling(old, new)
             )
             Popup.Message(tmp)
             pth = svtmp + pth
-          end
-          if Builtins.contains(SubvolNames(new), pth)
+          elsif Builtins.contains(SubvolNames(new), pth)
```
imobach: This PR fixes the original issue but it introduces a new one. If you try to add a `subvolume` which name does not start by *@/*, YaST will [warn the user](https://github.com/alanbortu/yast-storage/blob/4a3e6d1ba74a7b3d5a61daf1a6353f701b12dd0e/src/include/partitioning/custom_part_dialogs.rb#L1667) with this message: *Only subvolume names starting with "@/" currently allowed! Automatically prepending "@/" to name of subvolume.*

Unfortunately, the subvolume won't be created.

imobach: BTW, thanks for your patch. Could you check this issue?

alanbortu: Sure, I will take a look at it. Thank you for pointing out the issue, I missed it when I was testing my change :)

ancorgs: @alanbortu Any update on this?

alanbortu: Sorry, I just finished up my school exams today, I will take a look at the code now

alanbortu:
This seems to resolve the issue regarding the subvolume not being added if it did not start with "@". I moved the if statement inside the else so it only runs the check when its sure the subvolume name does not violate the other two requirements (non empty name and doesn't already exist). 

What it was doing before was checking if the name didn't start with "@" and if that was the case it skipped the creation.

alanbortu:
Okay, this commit should resolve the issues, I tested by trying the various cases the code checks for and they all seem to behave as expected. Please let me know if there was something I missed!

ancorgs:
@alanbortu please, create a new clean pull request containing just the fix and not all these meaningless commits and artifacts.

Other than that, I guess @imobach can take a look to the code, since he found the problems in the original commit.

alanbortu:
Ok will do, thanks @ancorgs

PR #173, opened 2015-11-26
Remove wrong change regarding bsc#940374

imobach:
This line was introduced in https://github.com/yast/yast-storage/pull/146 (reverted).


gabi2:
LGTM

PR #174, opened 2015-11-26
Revert "Merge pull request #169 from yast/netdev_fix_master"

gabi2:
This change is not wanted anymore as the root of the issue has been fixed in systemd already, see https://bugzilla.novell.com/show_bug.cgi?id=949683#c28

This reverts commit deefce05a32773247b13253d2daf6850800bc838, reversing
changes made to e5979fb5f30ec3d40544c0f21fd8bee84e7912de.

### TODO

- [ ] Fix version and changes

### DIFF
--- a/package/yast2-storage.changes
+++ b/package/yast2-storage.changes
@@ -1,9 +1,3 @@
-Wed Nov 11 14:21:32 CET 2015 - gs@suse.de
-
-- Add _netdev flag for iSCSI and FCoE backed fstab entries
-  (bsc#949683, patch provided by david.bond@microfocus.com)
-- 3.1.72
-
 -------------------------------------------------------------------
 Thu Oct 29 11:52:45 CET 2015 - shundhammer@suse.de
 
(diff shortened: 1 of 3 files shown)


lslezak:
Um, for me it's not clear why the commit is reverted. What's the reason for the revert?

gabi2:
see https://bugzilla.novell.com/show_bug.cgi?id=949683#c28

lslezak:
Ok, it's not needed as it was fixed in systemd...

The problem is that the package has been already submitted to Factory (https://build.opensuse.org/package/view_file/openSUSE:Factory/yast2-storage/yast2-storage.changes?expand=1). I guess version downgrade would make some troubles, `zypper up` would not install the newer package...

We cannot simply revert the commit. We have to **increase** the version number and **add** a new changelog entry describing the removal, the old entry should be kept...

kobliha:
I've updated the PR description.

gabi2:
You are right, the old changelog is needed and a new one about reverting and increase of version - done.

imobach:
LGTM

PR #175, opened 2015-11-30
New storage proposal based on new libstorage (tech preview)

shundhammer:


src/lib/storage/boot_requirements_checker.rb
```diff
@@ -0,0 +1,97 @@
...
+      def initialize(settings)
+        Yast.import "Arch"
+        @settings = settings
+      end
+
+      def needed_partitions
+        boot_volumes = []
+        boot_volumes << make_efi_boot_partition if efi_boot_partition_needed?
+        boot_volumes << make_boot_partition     if boot_partition_needed?
+        boot_volumes << make_prep_partition     if prep_partition_needed?
```
jreidinger: from code it looks like it make sense to have partition for all three cases, but in fact only one is needed, not?

src/lib/storage/disk_analyzer.rb
```diff
@@ -0,0 +1,368 @@
...
+          ::Storage::ID_LVM,
+          ::Storage::ID_RAID
+        ]
+
+      WINDOWS_PARTITION_IDS =
+        [
+          ::Storage::ID_NTFS,
+          ::Storage::ID_DOS32,
+          ::Storage::ID_DOS16,
+          ::Storage::ID_DOS12
```
jreidinger: do we really still support dos 12 and 16? I thought that it will be dropped also in kernel

src/lib/storage/disk_analyzer.rb
```diff
@@ -0,0 +1,368 @@
...
+        @linux_partitions   = []
+        @windows_partitions = [] # only filled if @linux_partitions is empty
+
+        # Maximum number of disks to check. This might be important on
+        # architectures that tend to have very many disks (s390).
+        @disk_check_limit = DEFAULT_DISK_CHECK_LIMIT
+      end
+
+      # Analyze disks and partitions. Make sure to call this before querying
+      # any member variables.
```
jreidinger: why not using lazy loading? in such case no analyze method needed.
with something like 
```
def installation_disks
  return @installation_disks if @installation_disks

  @installation_disks = find_installation_disks
end
```
it is less error-prone ( as you do not need to call anything before using it ) and easier to use.

src/lib/storage/disk_analyzer.rb
```diff
@@ -0,0 +1,368 @@
...
+          # Windows partitions in that case. This saves mounting and unmounting
+          # all Windows-like partitions.
+          log.info("Linux partitions found - not checking for Windows partitions")
+        end
+
+        log.info("Installation disks: #{dev_names(@installation_disks)}")
+        log.info("Candidate    disks: #{dev_names(@candidate_disks)}")
+        log.info("Linux   partitions: #{dev_names(@linux_partitions)}")
+        log.info("Windows partitions: #{dev_names(@windows_partitions)}")
+      end
```
jreidinger: I expect here access limitation like `private` as there is only few public methods. It really helps e.g. in generated documentation if you have only two public methods instead of ten.

src/lib/storage/disk_analyzer.rb
```diff
@@ -0,0 +1,368 @@
...
+
+      # Find any Linux partitions on any of the candidate disks.
+      # This may be a normal Linux partition (type 0x83), a Linux swap
+      # partition (type 0x82), an LVM partition, or a RAID partition.
+      #
+      # @return [Array<Partition>] Linux partitions
+      #
+      def find_linux_partitions
+        linux_partitions = []
+        @candidate_disks.each do |disk|
```
jreidinger: you can use 
```
linux_partitions = @candidate_disks.each_with_object([]) do |disk, linux_partitions|
```

src/lib/storage/disk_analyzer.rb
```diff
@@ -0,0 +1,368 @@
...
+          begin
+            disk.partition_table.partitions.each do |partition|
+              if LINUX_PARTITION_IDS.include?(partition.id)
+                log.info("Found Linux partition #{partition} (ID 0x#{partition.id.to_s(16)})")
+                linux_partitions << partition
+              end
+            end
+          rescue RuntimeError => ex
```
jreidinger: this is quite error prone as it catch all exception, even ones that are .e.g user fault or programmer errors, so in the end it will look like in old yast days, where errors are silently ignored and then in the end user find that something is wrong. I think that specialized error helps here.

src/lib/storage/disk_analyzer.rb
```diff
@@ -0,0 +1,368 @@
...
+      # Find disks that are suitable for installing Linux.
+      # Put any USB disks to the end of that array.
+      #
+      # @return [Array<::Storage::Disk>] candidate disks
+      #
+      def find_candidate_disks
+        disks = StorageManager.instance.probed.all_disks.to_a
+        # FIXME: to_a should not be necessary
+
+        usb_disks, non_usb_disks = disks.partition { |disk| disk.transport == ::Storage::USB }
```
jreidinger: nice :+1:

src/lib/storage/disk_analyzer.rb
```diff
@@ -0,0 +1,368 @@
...
+      # Remove any installation disks from 'disks' and return a disks array
+      # containing the disks that are not installation media.
+      #
+      # @param disks [Array<::Storage::Disk>]
+      # @return [Array<::Storage::Disk>] non-installation disks
+      #
+      def remove_installation_disks(disks)
+        # We can't simply use
+        #   disks -= @installation_disks
+        # because the list elements (from libstorage) don't provide a .hash method.
```
jreidinger: maybe it make sense to add it?

src/lib/storage/disk_analyzer.rb
```diff
@@ -0,0 +1,368 @@
...
+      end
+
+      # Return an array of the device names of the specified block devices
+      # (::Storage::Disk, ::Storage::Partition, ...).
+      #
+      # @param disks [Array<BlkDev>]
+      # @return [Array<string>] names
+      #
+      def dev_names(blk_devices)
+        blk_devices.map { |x| x.to_s }
```
jreidinger: can be written also like `blk_devices.map(&:to_s)`

src/lib/storage/disk_analyzer.rb
```diff
@@ -0,0 +1,368 @@
...
+      # Unmount a device.
+      #
+      # This is a temporary workaround until the new libstorage can handle that.
+      #
+      def umount(mount_point)
+        # FIXME: use libstorage function when available
+        cmd = "/usr/bin/umount #{mount_point}"
+        log.info("Unmounting: #{cmd}")
+        raise "umount failed for #{mount_point}" unless system(cmd)
+      end
```
jreidinger: I know that mount/umount is just temporary workaround, but in general for such pair of operation in ruby is encouraged to use block form with usage e.g.
```
mount(dev, mp) do
  FileUtils.identical?("#{mp}/a", my_a)
end
```

jreidinger: reason is e.g. when you raise exception between mount and umount, to properly umount it, otherwise there is risk of leaked mount points.

src/lib/storage/boot_requirements_checker.rb
```diff
@@ -0,0 +1,97 @@
...
+    # Class that can check requirements for the different kinds of boot
+    # partition: /boot, EFI-boot, PReP.
+    #
+    # TO DO: Check with arch maintainers if the requirements are correct.
+    #
+    # See also
+    # https://github.com/yast/yast-bootloader/blob/master/SUPPORTED_SCENARIOS.md
+    #
+    class BootRequirementsChecker
+      include Yast::Logger
```
ancorgs: NP: I don't see any logging in the existing code.

src/lib/storage/disk_analyzer.rb
```diff
@@ -0,0 +1,368 @@
...
+      # @return the return value of 'block' or 'nil' if there was an error.
+      #
+      def mount_and_check(vol, &block)
+        raise ArgumentError, "Code block required" unless block_given?
+        mount_point = "/mnt" # FIXME
+        begin
+          # check if we have a filesystem
+          # return false unless vol.filesystem
+          mount(vol.name, mount_point)
+          check_result = block.call(mount_point)
```
jreidinger: here is problem when block raise exception, then it is never unmount.

src/lib/storage/disk_size.rb
```diff
@@ -0,0 +1,260 @@
...
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "yast"
+require "pp"
```
jreidinger: is there reason for pretty print?

src/lib/storage/disk_size.rb
```diff
@@ -0,0 +1,260 @@
...
+          DiskSize.new(size * (1024**5))
+        end
+
+        def ZiB(size)
+          DiskSize.new(size * (1024**6))
+        end
+
+        def YiB(size)
+          DiskSize.new(size * (1024**7))
+        end
```
jreidinger: in fact in ruby you can do even more cool stuff like having refinements like https://github.com/yast/yast-ruby-bindings/blob/master/src/ruby/yast/core_ext/ansi_string.rb#L4
and explanation http://timelessrepo.com/refinements-in-ruby

jreidinger: using it, you can do things like 
```
using DiskSize
size = 15.GiB
```
 which looks better then `size = DiskSize.GiB(15)`

src/lib/storage/storage_manager.rb
```diff
@@ -0,0 +1,102 @@
...
+    #
+    # You can simply use StorageManager.instance to use it and create a
+    # libstorage instance if there isn't one yet; this will use common default
+    # parameters for creating it.
+    #
+    # If you need special parameters for creating it, you can use
+    # StorageManager.create_instance with a custom ::Storage::Environment
+    # (notice the difference between the global ::Storage namespace which is
+    # libstorage and this Yast::Storage namespace).
+    #
```
ancorgs: Note to self (will comment further after reading the whole code).

I'm still in the early stage of reading the code, but this usage of the singleton patterns looks slightly weird to me.
In general, I'm not a fan of the pattern, but making it possible to initialize the singleton object in different ways (with and without params), makes it worse.

src/lib/storage/disk_size.rb
```diff
@@ -0,0 +1,260 @@
...
+
+        def unlimited
+          DiskSize.new(-1)
+        end
+
+        def zero
+          DiskSize.new(0)
+        end
```
jreidinger: these two is in fact constants and not methods.

src/lib/storage/disk_size.rb
```diff
@@ -0,0 +1,260 @@
...
+        alias_method :from_s, :parse
+        alias_method :from_human_readable, :parse
+
+        # Return the unit exponent for any of the known binary units ("kiB",
+        # "MiB", ...). The base of this exponent is 1024. The base unit is kiB.
+        #
+        def unit_exponent(unit)
+          index = UNITS.index(unit)
+          raise ArgumentError, "expected one of #{UNITS}" if index.nil?
+          index
```
jreidinger: in fact in ruby it you can write it using work-flow or
```
def unit_exponent(unit)
  UNITS.index(unit) or raise ArgumentError, "expected one of #{UNITS}"
end
```

src/lib/storage/disk_size.rb
```diff
@@ -0,0 +1,260 @@
...
+      # Operators
+      #
+
+      def +(other)
+        if other.is_a?(Numeric)
+          DiskSize.new(@size_k + other)
+        elsif other.respond_to?(:size_k)
+          DiskSize.new(@size_k + other.size_k)
+        else
+          raise TypeError, "Numeric value or DiskSize expected"
```
jreidinger: better to specify class of passed value like `raise TypeError, "Unexpected #{other.class}, expected Numeric value or DiskSize"`

src/lib/storage/disk_size.rb
```diff
@@ -0,0 +1,260 @@
...
+      # Other methods
+      #
+
+      def unlimited?
+        size_k == -1
+      end
+
+      def zero?
+        size_k == 0
+      end
```
jreidinger: nice methods :+1:

src/lib/storage/disk_size.rb
```diff
@@ -0,0 +1,260 @@
...
+
+      # The Comparable mixin will get us operators < > <= >= == != with this
+      def <=>(other)
+        if other.respond_to?(:size_k)
+          return @size_k <=> other.size_k
```
jreidinger: I think there is one logical error. Using -1 for unlimited mean, that unlimited is smaller then zero, which is from mathematical point of view invalid.

src/lib/storage/disk_size.rb
```diff
@@ -0,0 +1,260 @@
...
+        return [UNLIMITED, ""] if size_k == -1
+
+        unit_index = 0
+        size = @size_k.to_f
+
+        while size > 1024.0 && unit_index < UNITS.size - 1
+          size /= 1024.0
+          unit_index += 1
+        end
+        [size, UNITS[unit_index]]
```
jreidinger: hint not translatable, so at least add FIXME

src/lib/storage/boot_requirements_checker.rb
aschnell: Sometimes two are needed, e.g. PReP and /boot (https://bugzilla.suse.com/show_bug.cgi?id=940465).

src/lib/storage/proposal_settings.rb
```diff
@@ -0,0 +1,93 @@
...
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "yast"
+require_relative "./disk_size"
```
jreidinger: it is discouraged technique. It should be part of common ruby load path. Using this you basically breaks e.g. using y2update in installation, if you want to patch only one file.
it should be 
```
  require "storage/disk_size"
```

jreidinger: that brings me to idea if it is good to place it into storage path...maybe use ystorage path?

src/lib/storage/proposal_settings.rb
```diff
@@ -0,0 +1,93 @@
...
+      def initialize
+        @use_lvm                  = false
+        @encrypt_volume_group     = false
+        @root_filesystem_type     = ::Storage::BTRFS
+        @use_snapshots            = true
+        @use_separate_home        = true
+        @home_filesystem_type     = ::Storage::XFS
+        @enlarge_swap_for_suspend = false
+      end
+    end
```
jreidinger: I think it make sense to consider using Struct class as this is mainly structure, not class with methods. see http://ruby-doc.org/core-2.2.0/Struct.html

src/lib/storage/proposal_volume.rb
```diff
@@ -0,0 +1,59 @@
...
+
+require "yast"
+require "pp"
+
+module Yast
+  module Storage
+    # Class to represent a planned volume (partition or logical volume) and
+    # its constraints
+    #
+    class ProposalVolume
```
jreidinger: also looks like struct for me ( a lot of data and almost no methods on it )

src/lib/storage/space_maker.rb
```diff
@@ -0,0 +1,76 @@
...
+    # or by resizing an existing Windows partition.
+    #
+    class SpaceMaker
+      include Yast::Logger
+
+      attr_reader :volumes
+
+      # Initialize.
+      #
+      # @param volumes [list of ProposalVolume] volumes to find space for.
```
jreidinger: usual notation in yardoc is `Array<ProposalVolume>`

src/lib/storage/space_maker.rb
```diff
@@ -0,0 +1,76 @@
...
+      # The volumes might be changed by this class.
+      #
+      # @param settings [Storage::Settings] parameters to use
+      #
+      # @param disk_analyzer [DiskAnalyzer]
+      #
+      def initialize(volumes, settings, disk_analyzer)
```
jreidinger: OK, disk analyzer is dependency injection. It is good, as it allows replacement. But it is good habit to provide default, so it is not mandatory to specify it. For me nicer API is to use defaults like
```
  def initialize(volumes, settings: StorageSettings.new, disk_analyzer: DiskAnalyzer.new)

  # and usage
  SpaceMaker.new(volumes)
  SpaceMaker.new(volumes, disk_analyzer: my_own_cool_analyzer)
  SpaceMaker.new(volumes, settings: user_settings || control_file_settings)
```

src/lib/storage/storage_manager.rb
```diff
@@ -0,0 +1,102 @@
...
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "yast"
+require "yastx"
```
jreidinger: explicit requirement for yastx should not be need, what is reason?

src/lib/storage/storage_manager.rb
```diff
@@ -0,0 +1,102 @@
...
+    # instance will also start hardware probing. This is why there is an alias
+    # name StorageManager.start_probing to make this explicit.
+    #
+    class StorageManager
+      include Yast::Logger
+      #
+      # Class methods
+      #
+      class << self
+        @instance = nil
```
jreidinger: this looks like singleton. Please use singleton builtin: http://ruby-doc.org/stdlib-1.9.3/libdoc/singleton/rdoc/Singleton.html

src/lib/storage/storage_proposal.rb
```diff
@@ -0,0 +1,138 @@
...
+      def initialize
+        @settings = ProposalSettings.new
+        @proposal = nil # ::Storage::DeviceGraph
+        @disk_blacklist = []
+        @disk_greylist  = []
+      end
+
+      # Create a storage proposal.
+      def propose
+        storage = StorageManager.instance # this will start probing in the first invocation
```
jreidinger: this is a bit tricky, what if something else already instance it? I think if you want explicit probe, you should call it.

src/lib/storage/disk_analyzer.rb
```diff
@@ -0,0 +1,368 @@
...
+
+        disks.each do |disk|
+          installation_disks << disk if installation_disk?(disk)
+          disk_count += 1
+          if disk_count > @disk_check_limit
+            log.info("Disk check limit reached after #{disk}")
+            break
+          end
+        end
+        installation_disks
```
ancorgs: The whole method looks not very Ruby-like. Nothing that will kill us :wink:, but what about?
```ruby
def find_installation_disks
  disks = StorageManager.instance.probed.all_disks.to_a
  # Check usb disks first
  disks.sort_by! {|disk| disk.transport == ::Storage::USB ? -1 : 1 }

  if disks.size > @disk_check_limit
    log.info("Too many disks, only the first {@disk_check_limit} ones will be cheched")
  end
  disks[0,@disk_check_limit].select {|disk| installation_disk?(disk)}
end
```

Not fully checked (maybe some `>` should be `>=` and that kind of things, but you get the idea

test/disk_size_test.rb
```diff
@@ -0,0 +1,211 @@
+#!/usr/bin/env rspec
+
+require_relative "spec_helper"
+require "storage/disk_size"
+require "pp"
+
+
+describe Yast::Storage::DiskSize do
+
+  describe "constructed empty" do
```
jreidinger: usual rspec structure is 
```
describe Yast::Storage::DiskSize do
  describe ".new" do
      context "no parameter passed" do
        it "has size_k 0" do
        end
      end
      
    context "integer passed" do
       it "has size_k equal integer" do
         ....
       end
     end
  describe ".KiB" do
    ....
```

idea of rspec is to have specification that is tested. Try use `rspec --color --format documentation` and output should give you API specification.

src/lib/storage/boot_requirements_checker.rb
jreidinger: ok, then it makes sense.

src/lib/storage/disk_analyzer.rb
ancorgs: I was going to suggest the same

src/lib/storage/disk_size.rb
jreidinger: It often helps, e.g. when you see NilClass you know that something try to add nil

src/lib/storage/disk_analyzer.rb
jreidinger: Another advantage is seen below. You can have default ctor passed as default. so for disk_analyzer you can by default pass as dependency injection new instance, and you are sure it will initialize when needed.

src/lib/storage/disk_analyzer.rb
```diff
@@ -0,0 +1,368 @@
...
+        @linux_partitions   = find_linux_partitions
+
+        if @linux_partitions.empty?
+          @windows_partitions = find_windows_partitions
+        else
+          # We only want to resize any MS Windows partition if there is no
+          # Linux already on that machine, so we don't need to check for
+          # Windows partitions in that case. This saves mounting and unmounting
+          # all Windows-like partitions.
+          log.info("Linux partitions found - not checking for Windows partitions")
```
ancorgs: Another advantage of lazy loading, it's the caller who decides if he wants/needs to check for windows partitions.

jreidinger: agreed, as it can be confusing to ask for windows partition and getting none, it can mean, we do not have any

src/lib/storage/disk_analyzer.rb
shundhammer: I don't know, but better be safe than sorry. It doesn't hurt us to have more of them. It's only for checking if there might be a windows partition on any of them.

src/lib/storage/disk_analyzer.rb
shundhammer: It doesn't make sense to do lazy init if we need it in 100% of all cases anyway - like here.
Also, I want to be in control when possibly dangerous operations happen - like mounting disks to check for certain stuff (doesn't happen right here and now, but in similar cases like windows_disks).

src/lib/storage/disk_analyzer.rb
shundhammer: Well, yes; that would get rid of the first line of that function - and of code readability.

src/lib/storage/disk_analyzer.rb
shundhammer: Not quite. At first, I had a rescue clause of all Exceptions - which rubocop complained about, and rightfully so. Now it's only RuntimeError - which does not include most Ruby errors.

I'd very much prefer to catch only ::Storage::Exception here, but unfortunately the SWIG bindings swallow that base class of all libstorage exceptions - it derives them all just from RuntimeError, which is indeed bad. The real fix would be in the libstorage SWIG bindings.

I'll ad a FIXME for that. Right now, we can't do any better than this.

src/lib/storage/disk_analyzer.rb
shundhammer: Yes, if you know how to do that?

IIRC Arvin successfully expanded that ::Storage class with a Ruby extension. Not sure exactly what is needed for that, though.

src/lib/storage/disk_analyzer.rb
shundhammer: I know; rubocop told me so. But I consider this much better readable.

src/lib/storage/disk_analyzer.rb
shundhammer: That's what the mount_and_check method in line #319 does. It just uses those two methods as helpers.

src/lib/storage/boot_requirements_checker.rb
shundhammer: Yes. It also doesn't do much yet. So far, it's just very primitive stubs.

I fear that code will grow very quickly, and then there will be lots and lots of logging. ;-)

src/lib/storage/disk_size.rb
shundhammer: I use it for debugging on and off. It's a PITA to always have to add that require just for debug-printing one value.

src/lib/storage/storage_manager.rb
shundhammer: just for the record - from IRC #yast:

[13:34:09] <ancorgs> HuHa: most relevant things commented by jreidinger already. I agree in all his suggestions. After fixing all that, the only thing that would remain "smelly" to me is the StorageManager class, I'll try to come up with a more constructive idea for it
[13:34:47] <HuHa> ancorgs: arvin and I had talked a lot about this
[13:35:02] <HuHa> this is the thing that makes the whole thing usable
[13:35:25] <HuHa> because only with a singleton you don't have to pass that storage object around as a parameter everywhere
[13:35:41] <HuHa> it might be used in other parts of the code, too
[13:35:49] <HuHa> and we certainly don't want to do the probing several times
[13:36:04] <HuHa> and we need to be able to set it up differently to fake storage things
[13:36:35] <HuHa> so we need to be able to initialize it from someplace else with a different ::Storage::Environment
[13:38:35] <ancorgs> HuHa: I see. I guess I need to read a little bit about ::Storage::Environment first :-)
[13:39:19] <HuHa> ancorgs: the nice thing about this approach is that you don't need to do exactly that ;-)
[13:39:23] <arvin> please review https://github.com/openSUSE/libstorage-bgl-eval/pull/25
[13:39:33] <HuHa> only if you have special requirements, you need to take care of that Environment
[13:39:37] <HuHa> in most cases, you don't
[13:41:10] <ancorgs> HuHa: well, maybe I don't need to do it as a class user, but I'll need to do it as a class code-reviewer :-)
[13:41:27] <HuHa> that's why we are talking to each other ;-)

src/lib/storage/disk_size.rb
shundhammer: Good idea. Changed accordingly.

src/lib/storage/disk_size.rb
shundhammer: Changed in all 5 locations.

src/lib/storage/disk_size.rb
shundhammer: Right, special treatment for unlimited? is missing. Will add.

src/lib/storage/disk_size.rb
shundhammer: Added.

src/lib/storage/proposal_settings.rb
shundhammer: I tried several things, but it always failed in one case or the other. I am using require_relative because I use the small piece of code at the end of the file for minimalistic testing. When I do that, it won't use the normal load path because I have a different working directory. I don't have a better solution for that right now. Do you?

src/lib/storage/proposal_settings.rb
shundhammer: It will get at least its read_from_xml method. That doesn't make it much of a class, but still it has a few methods.

src/lib/storage/space_maker.rb
shundhammer: Changed.

src/lib/storage/storage_manager.rb
shundhammer: No, it's different. Look closely.

We need a possibility to supply a different ::Storage::Environment to be able to do mocking for libstorage and other parameters that Environment may contain. 

Also, there were plans to do a clean shutdown for libstorage (right now difficult with Ruby's garbage collection, thus postponed).

src/lib/storage/storage_proposal.rb
shundhammer: If somebody else already triggered the probing, that's fine. It will probe with the first usage, and that's just what is desired here. We certainly don't want to probe several times. This comment is supposed to mean "this will start probing unless somebody else already did that someplace else".

src/lib/storage/disk_analyzer.rb
shundhammer: Yikes. That's what people hate about Ruby. Why change perfectly well readable code into gibberish without a pressing need to do so?

It begins with the sorting: I could never tell which stuff gets sorted first. The old storage proposal code is full of things like that; you'd have to tear it apart, add logging and execute it to find out what the hell this does. OTOH that

    usb_disks, non_usb_disks = disks.partition { |disk| disk.transport == ::Storage::USB }
    disks = usb_disks + non_usb_disks

is self-explanatory IMHO.

Also, that changed limit check and its logging will not tell me what disks were considered and which ones weren't - since it doesn't output even the device names. But then, you are right, that output is missing in my version as well. I'll change that.

src/lib/storage/disk_analyzer.rb
jreidinger: well, it is not 100% you already sneak in logic to not analyze windows if there is linux partitions, which is wrong from my POV as it should not be decision of this class. Be in control also means, that you can have potential problems like below if you forgot to pass already analyzed object, it will return empty arrays, which looks even more wrong for me, as it is not immediatell obvious what is problem.
What is reason to have control over mounting? for me it is nature to get exception if I call windows_partitions if something wrong happen when trying to detect it, so I know that information is unreliable from that method ( as opposed to current situation when it can be empty if I forgot analyze or only partial if some mount failed ).

src/lib/storage/disk_analyzer.rb
jreidinger: another way is `reduce`, but it require to always return from block reduced result. Not strong requirement, just helps with variable scopes.

src/lib/storage/disk_analyzer.rb
jreidinger: For me the biggest problem with it, is that you silently return only partial result and API user do not see it, which is wrong from my POV. E.g. consider I would like to do some decision based on linux_partitions, if it is not complete, then my decision will be wrong.

jreidinger: Consider it if this is your salary and function count number of your working days. If some day contain problem ( like you fill something wrong ), then you do not get salary for that day without any error message. At least me it cause very unhappy and I really prefer to know, that there is problem with counting my working days rather then silently decrease my salary.

src/lib/storage/disk_analyzer.rb
jreidinger: usual way how to count hash is to call hash on combination of unique identifiers. So if disk is identified by its name then you just need something like `def_delegators :hash, @name`
or if it is e.g. combination of its graph and name then use something like
```
  def hash
    [@name, @graph.id]. hash
  end
```

src/lib/storage/disk_analyzer.rb
jreidinger: it depends on background, if you get familiar with it, it really looks nature.
Compare how you read code:
- your one: "Map each blk_device to x and use result of call to_s on x" . Not bad, quite short, just mapping is not necessary

- suggestion: "Map each blk to its to_s result". Shorter and remove unnecessary mapping, so less brain capacity needed t do mapping.

src/lib/storage/disk_size.rb
shundhammer: Added special treatment for unlimited? for both operands (self and other) and testcases for each combination.

src/lib/storage/storage_manager.rb
ancorgs: Yes, it's different. And that's why I find this class so confusing. It pretends to be a Singleton class, but it's not. Why?
```ruby
RealSingletonClass.instance.class #=> RealSingleton
StorageManager.instance.class #=> Storage::Storage
```

ancorgs: I'll propose an alternative I'm thinking of. To be honest, I don't love it but I find it less confusing.

ancorgs: One alternative idea (not tested). Not nice but hopefully less confusing.
```ruby
module Yast
  module Storage
    module Libstorage
      @storage = nil
      @logger   = nil

      # @return [::Storage::Storage] libstorage object
      def self.storage
        create_storage unless @storage
        @storage
      end

      # Provide your own Storage::Environment for custom purposes like mocking
      # the hardware probing etc.
      #
      # @return [::Storage::Storage] libstorage object
      def self.create_storage(storage_environment: ::Storage::Environment.new(true))
        create_logger
        log.info("Creating Storage object")
        @storage = ::Storage::Storage.new(storage_environment)
      end

      private

      def self.create_logger
        @logger = StorageLogger.new
        ::Storage.logger = @logger
      end
    end
  end
end
```

shundhammer: Just think about how you'd use that thing:

    disks = Libstorage.storage.probed.all_disks

For one thing, this unnecessarily exposes that Libstorage is an external library. For another, this repetition of 'storage' gets confusing, even more so in a module Yast::Storage. I don't like either aspect of that.

ancorgs: Sure I suck at naming things. `Libstorage` is only one (bad) option. I called it `NeedsAGoodName` while writing the code, if you prefer. :wink: But from a practical point of view
```
disks = Libstorage.storage.probed.all_disks
```
is not more complex than the current.
```
disks = StorageManager.instance.probed.all_disks
```

On the other hand, both approaches return a libstorage object, so the external library is completely exposed from the very beginning. You'll need to read the libstorage documentation in order to use the object you are getting, so making the "externality" even more obvious doesn't look like a problem to me.

We could, of course, go even one step further with this approach and do this nasty thing. But I wouldn't recommend it because it would be using the ```Yast::Storage``` module for two things at the same time (namespacing and providing methods).
```ruby
module Yast
  module Storage
    @libstorage_instance = nil
    @libstorage_logger   = nil
        
    # @return [::Storage::Storage] libstorage object
    def self.libstorage_instance
      create_libstorage_instance unless @storage_instance
      @libstorage_instance
    end
        
    # Provide your own Storage::Environment for custom purposes like mocking
    # the hardware probing etc.
    #
    # @return [::Storage::Storage] libstorage object
    def self.create_libstorage_instance(storage_environment: ::Storage::Environment.new(true))
      create_libstorage_logger
      log.info("Creating Storage object")
      @storage = ::Storage::Storage.new(storage_environment)
    end 
      
    private
      
    def self.create_logger
      @logger = StorageLogger.new
      ::Storage.logger = @logger
    end
  end
end
```
Usage would be
```ruby
disk = libstorage_instance.probed.all_disks
```

src/lib/storage/proposal_settings.rb
jreidinger: proper way is to set Y2DIR like in common usage. ( e.g. in test cases we do it this way )

src/lib/storage/proposal_settings.rb
jreidinger: see example in link I posted. Of course it can have methods. Just it simplify creation of many data fields.

src/lib/storage/proposal_settings.rb
jreidinger: btw nice shortcut if you want to run client from git checkout is `rake run[clientname]`

src/lib/storage/storage_proposal.rb
jreidinger: ok, so lazy loading, fine for me. Not sure if it need specific comment, but no problem for me

src/lib/storage/proposal_volume.rb
mvidner: "Proposal Volume" does not mean "Proposed Volume" or "Proposal for a Volume". We should make a better name.

src/lib/storage/proposal_volume.rb
```diff
@@ -0,0 +1,59 @@
...
+      attr_accessor :mount_point, :filesystem_type
```
mvidner: Please annotate them with types: I can see that `mount_point` appears to be a plain String, `*_size` are DiskSize, but it will be good to link to the `filesystem_type` enum (?) etc

shundhammer: Right. filesystem_type is a libstorage enum. Will add documentation.

shundhammer: Done.

src/lib/storage/disk_size.rb
shundhammer: I just tried converting them to constants; then I can non longer write

    my_size1 = DiskSize.GiB(200)
    my_size2 = DiskSize.unlimited
    my_size3 = DiskSize.TiB(2)
    my_size4 = DiskSize.zero

I have to write

    my_size1 = DiskSize.GiB(200)
    my_size2 = DiskSize::UNLIMITED
    my_size3 = DiskSize.TiB(2)
    my_size4 = DiskSize::ZERO

The inconsistency between . and :: is yet another unnecessary cause for errors. In Ruby that's a lot worse because they only get detected during runtime.

src/lib/storage/storage_manager.rb
shundhammer: Removed.

I thought I had needed it, but I don't remember why. Removing it didn't do any harm.

src/lib/storage/disk_size.rb
jreidinger: ok, so at least you can use it internally, as DiskSize for me looks immutable, then it make sense to not create always new instance when calling this method and instead used shared instance for unlimited and zero.

src/lib/storage/storage_manager.rb
jreidinger: I am curious because yast require it, so maybe in past you do not require complete yast module, but only its parts and then it can be useful to require yast parts written in C.

PR #176, opened 2015-12-04
Fixes to allow creating a LVM-based proposal based on control file settings

jsrain:


lslezak:
Looks OK to me, but I'm not a storage expert...

src/modules/StorageProposal.rb
```diff
@@ -6116,6 +6116,9 @@ def get_proposal_vm(target, key, disk)
 
 
     def get_inst_prop(target)
+      # initialize data from control file earlier, it is needed in this function
+      # to decide whteher to use LVM proposal (bsc#957913)
```
aschnell: typo 'whether'

aschnell:
LGTM

Unfortunately the explicit StorageProposal.GetControlCfg() call in helper.rb in the testsuite cannot be removed since some test cases modify StorageProposal settings that cannot be specified in ProductFeatures (e.g. empty9.rb).

Issue #178, opened 2015-12-14
get_inst_proposal does not honor default_fs


e9925248:
src/modules/StorageProposal.rb function get_inst_proposal
```
      root = {
        "mount"       => "/",
        "increasable" => true,
        "fsys"        => PropDefaultFs(),
        "size"        => 0
      }
      opts = GetControlCfg()
```

The call to GetControlCfg needs to be moved before setting up the root structure:
PropDefaultFs() returns proposal_root_fs, which has at that point the default value btrfs - so the proposal is always based on btrfs.
GetControlCfg calls SetProposalDefault, which in turn calls SetProposalRootFs to initialise the root fs type from /etc/sysconfig/storage.  Switching both commands would make sure, that the proposal honors the user settings.

ancorgs:
Please, use openSUSE's bugzilla to report and track errors. http://bugzilla.opensuse.org

e9925248:
https://bugzilla.opensuse.org/show_bug.cgi?id=959435

ancorgs:
@e9925248 Good point. We wrote that before bugzilla/fate reports started to be mandatory. We'll probably need to fix contributing.md. Noted down.

ancorgs:
Closed by #179 (already merged and sent to TW). More info in the corresponding bugzilla entry.

PR #179, opened 2015-12-16
Live installer should honor default fs

e9925248:
See https://github.com/yast/yast-storage/issues/178

The call to GetControlCfg needs to be moved before setting up the root structure:
PropDefaultFs() returns proposal_root_fs, which has at that point the default value btrfs - so the proposal is always based on btrfs.
GetControlCfg calls SetProposalDefault, which in turn calls SetProposalRootFs to initialise the root fs type from /etc/sysconfig/storage. Switching both commands would make sure, that the proposal honors the user settings.

### DIFF
--- a/src/modules/StorageProposal.rb
+++ b/src/modules/StorageProposal.rb
@@ -4503,13 +4503,13 @@ module Yast
       ret = {}
       target = AddWinInfo(target)
       Ops.set(ret, "target", target)
+      opts = GetControlCfg()
       root = {
         "mount"       => "/",
         "increasable" => true,
         "fsys"        => PropDefaultFs(),
         "size"        => 0
       }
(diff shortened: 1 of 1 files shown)


shundhammer:
LGTM

ancorgs:
Cool, if we want this in next tumbleweed snapshot we'll probably need a bugzilla entry (see comment at #178), so we can update the changelog and the version number, since lately a bugzilla entry is usually requested for every change.

@e9925248 do you want me to take care?

ancorgs:
Ok, now that we have the bugzilla entry (thanks @e9925248) and the LGTM (thanks @shundhammer). I'll merge this and create a separate PR for the changelog entry, so it goes into factory ASAP.

PR #180, opened 2015-12-16
Fix boo#955103

alanbortu:
As discussed in https://github.com/yast/yast-storage/pull/170 making new request to get rid of the noise commits.

Resolves issue regarding YaST2 partitioning module allowing you to enter an empty subvolume name.

### DIFF
--- a/src/include/partitioning/custom_part_dialogs.rb
+++ b/src/include/partitioning/custom_part_dialogs.rb
@@ -1664,21 +1664,22 @@ module Yast
           Builtins.y2milestone("SubvolHandling names:%1", SubvolNames(new))
           if pth == nil || Builtins.size(pth) == 0
             Popup.Message(_("Empty subvolume name not allowed."))
-          elsif Ops.greater_than(Builtins.size(FileSystems.default_subvol), 0) &&
-              Builtins.substring(pth, 0, Builtins.size(svtmp)) != svtmp
-            tmp = Builtins.sformat(
-              _(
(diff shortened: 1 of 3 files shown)


src/include/partitioning/custom_part_dialogs.rb
```diff
@@ -1664,31 +1664,37 @@ def SubvolHandling(old, new)
...
+                  "Only subvolume names starting with \"%1\" currently allowed!\nAutomatically prepending \"%1\" to name of subvolume."
+                ),
+                svtmp
               )
-            )
-            changed = true
+              Popup.Message(tmp)
+              pth = svtmp + pth  
+            end
+            if Builtins.contains(SubvolNames(new), pth)
```
imobach: This code looks like duplicated. I think that you could remove the branch starting in L#1667 and it should work (but I've not tried it). What do you think?

imobach:
Thanks @alanbortu!

For me it looks good. Anyway, I'd like that @shundhammer or @aschnell take a look.

On the other hand, maybe this PR should be against `SLE-12-SP1` as it's supposed to fix a bug in openSUSE Leap 42.1 and SLE 12 and Leap should share the same YaST version.

imobach:
After discussing in the IRC channel, and given that the bug is not critical nor important, we'll fix it only in master /cc @ancorgs

ancorgs:
Code looks good to me as well. I will merge it to master manually (I plan to break mergeability with another update, but I'll clean my own mess).

PR #182, opened 2015-12-19
Add option to disable swap creation

e9925248:
There is currently no option to disable the creation of swap partitions. 
While swap is necessary for some kinds of systems, swap can cause negative effects for other setups.

### DIFF
--- a/src/modules/StorageProposal.rb
+++ b/src/modules/StorageProposal.rb
@@ -65,6 +65,7 @@ module Yast
 
       @proposal_home = false
       @proposal_home_fs = :xfs
+      @proposal_swap = true
       @proposal_lvm = false
       @proposal_encrypt = false
       @proposal_root_fs = :btrfs
@@ -142,6 +143,14 @@ module Yast
       Builtins.y2milestone("SetProposalSnapshots val: %1", val)
     end
 
+    def GetProposalSwap()
+      @proposal_swap
+    end
+
+    def SetProposalSwap(val)
+      @proposal_swap = val
+      Builtins.y2milestone("SetProposalSwap val: %1", val)
+    end
 
     def GetProposalSuspend
       @proposal_suspend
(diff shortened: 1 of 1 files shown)


src/modules/StorageProposal.rb
```diff
@@ -289,6 +300,10 @@ def GetControlCfg
         tmp = ProductFeatures.GetFeature("partitioning", "proposal_snapshots")
         Ops.set(@cfg_xml, "prop_snapshots", tmp == "" || tmp == true ? true : false)
 
+        # GetBooleanFeature cannot distinguish between missing and false
+        tmp = ProductFeatures.GetFeature("partitioning", "need_swap")
+        Ops.set(@cfg_xml, "need_swap", tmp == "" || tmp == true ? true : false)
```
lslezak: Note: If you add support for a completely new option you need to define it in the [schema file](https://github.com/yast/yast-installation-control/blob/master/control/control.rnc#L440) otherwise the control files will not validate.

lslezak:
Is there any feature or bugzilla bug which you could reference?

shundhammer:
IMHO this is a non-feature: Only product creators who write control.xml files might even theoretically benefit from this, if at all. I don't recall a feature request or bug report requesting anything like this.

And even then,  this would only affect the automatic storage proposal which the user can always override. While some users might find it beneficial to have no swap at all (which is a disputed topic to say the least), most users will defer to the expertise of the distribution maker, which is us, and just accept the storage proposal.

I don't see any scenario where this patch would be useful. If one of our distribution media makers (who are the people who create the control.xml file) wanted this, this would be another matter; but this is not useful at all for the outside world. This would just introduce more complexity to test against and to maintain.

I vote for not accepting this change.

e9925248:
There is no bug report/feature request.

It was written for https://build.opensuse.org/package/show/home:e9925248:grandorgue/GOtoGO - A Live CD including the Live installer just to run one program.

Suspend/resume is not of interest, so swap is not necessary for this. 

My users try to stuff as much as possible and the program needs to keep all data fast accessible in RAM, because delays caused by swap-in from slow disks cause audio issues.

So the CD needs to present the user a proposal without swap.

shundhammer:
You might want to consider tuning the _swappiness_ rather than completely removing the option to have swap.

  https://en.wikipedia.org/wiki/Swappiness

Not having swap space at all is not desirable in the general case IMHO.

lslezak:
> My users try to stuff as much as possible and the program needs to keep all data fast accessible in RAM

But having no swap means that instead of seeing delays the kernel will kill some processes and users will see random crashes. And that's worse IMHO.

Stefan is right, if you want to minimize swapping fine tune the swappiness kernel parameter instead of not having swap at all. This will handle the critical situations better.

Please also note that the [YaST live installer](https://github.com/yast/yast-live-installer) has been dropped and is not supported anymore. (The openSUSE Tumbleweed Live medium runs the full installation, just like normal installation media.)

PR #183, opened 2015-12-21
Tech preview: New proposal based on new libstorage-bgl (part 2)

shundhammer:
This is not complete according to the PBI yet.

(Major) changes to the last version:

- Now using device names wherever possible, not C++ -meets-Ruby-objects from libstorage
  (this should be easier to test in an isolated way)

- Use libstorage-bgl to find unpartitioned space

- Actually change things using libstorage-bgl calls:
 Right now, the candidate disk is flattened (destroying all partitions and the partition table) for demonstration purposes.

- Introduced an intermediate libstorage-bgl devicegraph "proposal-base" to handle scenarios where the user selects a disk or partitions to be deleted to make room for the Linux installation (available now, too); that would be a copy of the probed devicegraph with those partitions already deleted. The general idea is to revert to this devicegraph when the proposal goes wrong or when the user wants to change proposal settings (with or without LVM, with or without separate home, etc.).

- Using a devicegraph "proposal" that starts as a copy of "proposal-base" to calculate the proposal.

- Creating an actiongraph with libstorage-bgl: This is the diff between the probed devicegraph and the "proposal" devicegraph.

- Using libstorage-bgl to get a textual description of the actions that will be performed (if the user commits those changes when confirming the installation).

### DIFF
--- a/src/lib/storage/disk_analyzer.rb
+++ b/src/lib/storage/disk_analyzer.rb
@@ -293,7 +293,7 @@ module Yast
       def installation_volume?(vol_name)
         log.info("Checking if #{vol_name} is an installation volume")
         is_inst = mount_and_check(vol_name) { |mp| installation_volume_check(mp) }
-        log.info("#{vol_name} is installation medium") if is_inst
+        log.info("#{vol_name} is installation medium: #{is_inst}")
         is_inst
       end
 
--- a/src/lib/storage/disk_analyzer.rb
+++ b/src/lib/storage/disk_analyzer.rb
@@ -67,9 +67,9 @@ module Yast
       def initialize
         Yast.import "Arch"
 
-        @installation_disks = []
-        @candidate_disks    = []
-        @linux_partitions   = []
+        @installation_disks = [] # device names of installation media
+        @candidate_disks    = [] # device names of disks to install on
+        @linux_partitions   = [] # device names of existing Linux parititions
         @windows_partitions = [] # only filled if @linux_partitions is empty
 
         # Maximum number of disks to check. This might be important on
@@ -95,10 +95,10 @@ module Yast
           log.info("Linux partitions found - not checking for Windows partitions")
         end
 
-        log.info("Installation disks: #{dev_names(@installation_disks)}")
-        log.info("Candidate    disks: #{dev_names(@candidate_disks)}")
-        log.info("Linux   partitions: #{dev_names(@linux_partitions)}")
-        log.info("Windows partitions: #{dev_names(@windows_partitions)}")
+        log.info("Installation disks: #{@installation_disks}")
+        log.info("Candidate    disks: #{@candidate_disks}")
+        log.info("Linux   partitions: #{@linux_partitions}")
+        log.info("Windows partitions: #{@windows_partitions}")
       end
 
       # Find disks that look like the current installation medium
@@ -109,7 +109,7 @@ module Yast
       # checking if a disk is an installation disk involves mounting and
       # unmounting each partition on that disk.
       #
-      # @return [Array<::Storage::Disk>] installation disks
+      # @return [Array<String>] device names of installation disks
       #
       def find_installation_disks
         disks = StorageManager.instance.probed.all_disks.to_a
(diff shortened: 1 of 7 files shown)


src/lib/storage/disk_analyzer.rb
```diff
@@ -286,10 +286,14 @@ def installation_disk?(disk)
...
-        is_inst = mount_and_check(vol) { |mp| installation_volume_check(mp) }
-        log.info("#{vol} is installation medium") if is_inst
+      # @param vol_name [string] device name of the volume to check
+      #
+      # @return [bool] 'true' if the volume is an installation volume
+      #
+      def installation_volume?(vol_name)
+        log.info("Checking if #{vol_name} is an installation volume")
+        is_inst = mount_and_check(vol_name) { |mp| installation_volume_check(mp) }
+        log.info("#{vol_name} is installation medium") if is_inst
```
lslezak: I'd always log the result here. If you log "Checking if FOO is an installation volume", there should be always also the result.

Otherwise you have to check the source code to find out that if the result message is missing it is not an installation volume. Which is quite tricky...

shundhammer: There is a summary of all that later anyway:

2015-12-21 15:09:44 <1> morgul(26268) [Ruby] storage/disk_analyzer.rb:187 Linux part: ["/dev/sda5", "/dev/sda1", "/dev/sda3", "/dev/sda2"]
2015-12-21 15:09:44 <1> morgul(26268) [Ruby] storage/disk_analyzer.rb:95 Linux partitions found - not checking for Windows partitions
2015-12-21 15:09:44 <1> morgul(26268) [Ruby] storage/disk_analyzer.rb:98 Installation disks: []
2015-12-21 15:09:44 <1> morgul(26268) [Ruby] storage/disk_analyzer.rb:99 Candidate    disks: ["/dev/sda"]
2015-12-21 15:09:44 <1> morgul(26268) [Ruby] storage/disk_analyzer.rb:100 Linux   partitions: ["/dev/sda5", "/dev/sda1", "/dev/sda3", "/dev/sda2"]
2015-12-21 15:09:44 <1> morgul(26268) [Ruby] storage/disk_analyzer.rb:101 Windows partitions: []

src/lib/storage/disk_size.rb
```diff
@@ -130,6 +130,7 @@ def unit_multiplier(unit)
       #
 
       def +(other)
+        return DiskSize.unlimited if any_operand_unlimited?(other)
```
lslezak: What about using the builtin Ruby support for infinity? Could it help and make some operations simpler?

```Ruby
> irb
irb(main):001:0> Float::INFINITY + 5
=> Infinity
irb(main):002:0> Float::INFINITY - 10
=> Infinity
irb(main):003:0> Float::INFINITY / 10
=> Infinity
```

src/lib/storage/space_maker.rb
```diff
@@ -38,39 +37,133 @@ module Storage
...
       end
 
       # Resize an existing MS Windows partition to free up disk space.
+      #
       def resize_windows_partition(partition)
         # TO DO
         partition
       end
+
+      # Delete all partitions on a disk.
```
lslezak: According to the code it seems that an empty partition table is created. This should be documented here.

src/lib/storage/disk_size.rb
shundhammer: You mean, now that I implemented that on my own without knowing that the concept exists in Ruby? ;-)

Not sure if everything would work well with the standard Ruby Fixnum I am using here for size_k in the DiskSize when I add Float::INFINITY to the mix. It might or might not work as expected. There might be weird side effects. My implementation is now well tested - the test cases should cover all scenarios IMHO.

lslezak: Um, I didn't know that as well, but when I saw your approach I could not resist to type "Ruby infinity" into Google... :wink: 

If it works for you then you can keep it, I was just looking for some more standard way...

lslezak:
I'm not a storage expert, but LGTM...

(Note: maybe more smaller PRs would be better than one big, but sometimes it's hard...)

PR #185, opened 2016-02-01
New storage proposal

shundhammer:
(like just discussed on Jangouts with jsrain, ancorgs, imo, snwint)

### DIFF
--- a/src/lib/storage/free_disk_space.rb
+++ b/src/lib/storage/free_disk_space.rb
@@ -22,8 +22,6 @@
 # find current contact information at www.suse.com.
 
 require "yast"
-require_relative "./disk_size"
-require "pp"
 
 module Yast
   module Storage
--- a/src/lib/storage/free_disk_space.rb
+++ b/src/lib/storage/free_disk_space.rb
@@ -37,14 +37,29 @@ module Yast
 
       # Initialize.
       #
-      # @param disk_name [string]
+      # @param disk [::Storage::Disk]
       #
-      # @param region [::Storage::PartitionSlot]
+      # @param slot [::Storage::PartitionSlot]
       #
-      def initialize(disk_name, slot)
-	@disk_name = disk_name
+      def initialize(disk, slot)
+	@disk = disk
 	@slot = slot
-	@size = DiskSize.new(slot.region.to_kb(slot.region.length))
+      end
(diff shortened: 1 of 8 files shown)


src/lib/storage/free_disk_space.rb
```diff
@@ -0,0 +1,66 @@
...
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "yast"
+require_relative "./disk_size"
+require "pp"
```
imobach: Is it needed?

src/lib/storage/free_disk_space.rb
```diff
@@ -0,0 +1,66 @@
...
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "yast"
+require_relative "./disk_size"
```
imobach: You can write it as `require_relative "disk_size"`.

src/lib/storage/partition_creator.rb
```diff
@@ -0,0 +1,372 @@
...
+
+require "yast"
+require "fileutils"
+require_relative "./storage_manager"
+require_relative "./proposal_settings"
+require_relative "./proposal_volume"
+require_relative "./disk_size"
+require_relative "./space_maker"
+require_relative "./free_disk_space"
+require "pp"
```
imobach: Same as above.

src/lib/storage/free_disk_space.rb
shundhammer: Right, here it's not needed. It probably was for a while, but now not any more.

src/lib/storage/free_disk_space.rb
shundhammer: Fixed.

src/lib/storage/partition_creator.rb
shundhammer: Fixed.

PR #186, opened 2016-02-04
Expert partitioner only offers snapshots if supported (bsc#944252)

ancorgs:
Also detected and fixed bsc#954691

### DIFF
--- a/src/include/partitioning/custom_part_dialogs.rb
+++ b/src/include/partitioning/custom_part_dialogs.rb
@@ -1533,11 +1533,11 @@ module Yast
 
 
     # Dialog: Subvolume handling
-    # @parm old map with original partition
-    # @parm new map with changes filled in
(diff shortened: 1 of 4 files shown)


src/include/partitioning/custom_part_dialogs.rb
```diff
@@ -1582,7 +1582,7 @@ def SubvolHandling(old, new)
         )
       )
 
-      if Mode.installation()
+      if Mode.installation() && snapshots_supported?(new)
```
jreidinger: why it use new and not self?...ah, great variable name. Is possible to rename variable to something like old_partitioning and new_partitioning?

ancorgs: Good point. Let's play boy scouts!

jreidinger:
LGTM

PR #188, opened 2016-02-09
Code shuffling. Read below

ancorgs:
- TODO: some classes still lack proper documentation
- Only one real functional change: added LVM to the list of partitions that cannot be installation media
- No more require_relative, demo moved to a proper `proposal_demo` yast client
- New namespace `Storage::Proposal` including an exceptions hierarchy
- Redistributed responsabilities and fewer class dependencies
- Volume renamed to PlannedVolume
- No more input/output arguments, classes always return a copy of the devicegraph (not registered in StorageManager)


src/lib/storage/proposal/partition_creator.rb
```diff
@@ -0,0 +1,364 @@
...
+        include Yast::Logger
+
+        attr_accessor :settings
+
+        VOLUME_GROUP_SYSTEM = "system"
+        FIRST_LOGICAL_PARTITION_NUMBER = 5 # Number of the first logical partition (/dev/sdx5)
+
+        # Initialize.
+        #
+        # @param original_graph [::Storage::Devicegraph] initial devicegraph
```
shundhammer: "Initial devicegraph" is misleading. I wrote that code, and I couldn't make sense of this.

This is nothing like an "original graph", it's the device graph this object works on. It is changed by this class, and the changes are visible to the outside, so this is not anything like an "initial" or "original" device graph -- which would imply that this class copies it or whatever. This is not true.

This is the device graph to work on, as the original documentation said. Nothing more, nothing less. Let's not confuse people who read this.

src/lib/storage/proposal/partition_creator.rb
```diff
@@ -0,0 +1,364 @@
...
+        #
+        def total_free_size
+          devicegraph.available_size
+        end
+
+        # List of free slots in the devicegraph
+        #
+        # @return [Array<FreeDiskSpace>]
+        #
+        def free_spaces
```
shundhammer: "free_spaces" is not a good name, in particular because everywhere else in this class it's called "free_slot" when one of them is passed as a parameter. It's also too easy to confuse free_size and free_space or free_spaces. It happened to me while I was working on this, that's why I clearly distinguished between free_slot/free_slots and free_size. Please let's remain clear what is meant with which term.

src/lib/storage/proposal/partition_creator.rb
shundhammer: Disregard. I only read just now that it really does copy the device graph internally.

src/lib/storage/proposal/partition_creator.rb
```diff
@@ -0,0 +1,364 @@
...
+          @settings = settings
+        end
+
+        # Returns a copy of the original devicegraph in which all the needed
+        # partitions have been created.
+        #
+        # @param volumes [PlannedVolumesCollection] volumes to create
+        # @param target_size [Symbol] :desired or :min
+        # @return [::Storage::Devicegraph]
+        def run(volumes, target_size)
```
shundhammer: "run" is a bit too generic for my taste. I think we should come up with a better name.

src/lib/storage/planned_volumes_collection.rb
```diff
@@ -0,0 +1,67 @@
...
+
+require "yast"
+require "storage/planned_volume"
+
+module Yast
+  module Storage
+    # Set of PlannedVolume elements
+    #
+    # Implements Enumerable and provides some extra methods to query the set
+    class PlannedVolumesCollection
```
shundhammer: Good idea to move this out to a separate class. I had toyed with that idea, but not really done it for lack of extra functionality, but those few extra methods here really make it worthwhile.

src/lib/storage/storage_manager.rb
```diff
@@ -72,7 +68,7 @@ def create_instance(storage_environment = nil)
           storage_environment ||= ::Storage::Environment.new(true)
           create_logger
           log.info("Creating Storage object")
-          @instance = ::Storage::Storage.new(storage_environment)
+          ::Storage::Storage.new(storage_environment)
```
shundhammer: **This is wrong.** When calling create_instance from the outside to initialize libstorage with a custom storage environment (e.g. for setting up fake device trees), @instance is not initialized here, so the next call to instance() will create another Storage object, overwriting the previous one with the special environment.

src/lib/storage/proposal/space_maker.rb
```diff
@@ -0,0 +1,134 @@
...
+        end
+
+        # Returns a copy of the original devicegraph in which all needed
+        # operations to free the required size have been performed
+        #
+        # @raise Proposal::NoDiskSpaceError if there is no enough room
+        #
+        # @param required_size [DiskSize] required amount of available space
+        # @return [::Storage::Devicegraph]
+        def run(required_size)
```
shundhammer: Again: "run" is too generic here. Why replace perfectly speaking names with generic ones that don't tell anything at all? It's like calling the main function "doit()". I don't like this at all. This is not an improvement.

src/lib/storage/proposal/devicegraph_generator.rb
```diff
@@ -0,0 +1,106 @@
...
+require "storage/proposal/partition_creator"
+
+module Yast
+  module Storage
+    module Proposal
+      using RefinedDevicegraph
+
+      # Class to create devicegraphs that can accommodate a given collection of
+      # volumes
+      class DevicegraphGenerator
```
shundhammer: Why throw away the speaking name StorageProposal and replace it with a nondescript name DevicegraphGenerator?

We are not in the business of making or collecting device graphs. This should be about making a useful storage proposal, and I don't see that name any more at all. This is a poor design choice. Whoever is reading this code will search forever to find anything that matches what he sees on the screen during installation, or well-known concepts in the YaST code.

I completely disagree with this choice.

src/lib/storage/proposal/space_maker.rb
```diff
@@ -0,0 +1,134 @@
...
+          graph.available_size >= required_size
+        end
+
+        # Try to resize an existing windows partition - unless there already is
+        # a Linux partition which means that
+        #
+        # @param devicegraph [DeviceGraph] devicegraph to update
+        # @param required_size [DiskSize]
+        def resize_windows!(devicegraph, required_size)
+          return if devicegraph.windows_part_names.empty?
```
shundhammer: Why did you remove @tried_resize_windows ? There was a good reason why this was there: To avoid resizing a windows partition over and over again as we try to make space for our Linux. How can we make sure now that this does not happen?

ancorgs: I don't remember. Maybe by accident :-) Let me check

src/lib/storage/proposal/devicegraph_generator.rb
ancorgs: The old proposal class is not gone. Is the new `ProposalDemoClient`. What I disliked about the old `Proposal` class is that it did quite some things: deciding the volumes, creating the target devicegraph, calculating the actions for that devicegraph, printing the result and all the intermediate steps if needed and even some additional tasks....

I just wanted to break that block into pieces: deciding the volumes goes now into a class that actually returns the list of volumes, you use those volumes to call a different class that calculates the final map, then you can add that map for the actions... and so on.

The old proposal class is still there. It's simply that at this point in time, it looks like a YaST client to me.

src/lib/storage/proposal/partition_creator.rb
ancorgs: I changed it from `free_space` to `free_spaces` since the first one doesn't suggest an array to me. But I have no strong feelings about it.

src/lib/storage/proposal/space_maker.rb
ancorgs: @shundhammer It's not needed anymore because now `SpaceMaker` has the responsibility and encapsulates the knowledge about how to free the space. There is nobody from outside telling `SpaceMaker` to resize windows at any random point in time.

shundhammer:
OK, much better now; this is what we agreed upon.

LGTM now.

PR #190, opened 2016-02-15
Include names in the shadowed subvolumes popup (bsc#928641)

ancorgs:


src/lib/storage/shadowed_vol_list.rb
```diff
@@ -0,0 +1,64 @@
...
+Yast.import "FileSystems"
+
+module Yast
+  class ShadowedVolList
+    include Enumerable
+    include Yast::Logger
+
+    attr_reader :partition, :mntpoint
+
+    def initialize(partition: nil, mntpoint: nil)
```
shundhammer: NP: mntpoint is not really that intuitive for a named parameter. Any good reason against mount_point which would be better to memorize?

src/lib/storage/shadowed_vol_list.rb
```diff
@@ -0,0 +1,64 @@
...
+      subvols = partition.fetch("subvol", [])
+      log.debug "Full list of subvolumes: #{subvols}"
+      subvols.select do |subvol|
+        name = full_name(subvol)
+        shadowed?(name)
+      end
+    end
+
+    def shadowed?(name)
+      (name == mntpoint) || (name.start_with?(mntpoint) && name[mntpoint.size] == "/")
```
shundhammer: Not sure what exactly this expression does:

   name[mntpoint.size] == "/"

maybe it deserves a dedicated function with a speaking name?

src/lib/storage/shadowed_vol_list.rb
```diff
@@ -0,0 +1,64 @@
...
+    end
+
+    def full_name(subvol)
+      if FileSystems.default_subvol.empty?
+        "/" + subvol.fetch("name")
```
shundhammer: subvol.fetch("name") is ycpkiller'ish for subvol["name"]?

src/include/partitioning/custom_part_check_generated.rb
```diff
@@ -549,6 +549,12 @@ def check_created_partition_table(targetMap, installation)
       ok
     end
 
+    def format_subvols(subvols)
+      # Extra space needed for the regexp
+      all = subvols.join(', ') + " "
```
shundhammer: Not sure why this lone space is added at the end when later there is a strip() call

shundhammer:
Aside from the small nitpicking issues LGTM.

src/lib/storage/shadowed_vol_list.rb
ancorgs: just typing laziness. I'll change it.

src/include/partitioning/custom_part_check_generated.rb
ancorgs: Read the comment :-)

src/lib/storage/shadowed_vol_list.rb
ancorgs: Not really ycpkiller'ish. I use fetch often also in new code when I consider it looks better :-)

src/lib/storage/shadowed_vol_list.rb
ancorgs: Good point. I'll do it

PR #191, opened 2016-02-16
Expert partitioner: userfrienly management of subvolumes (fate#320296)

ancorgs:
Tested manually.


src/lib/storage/shadowed_vol_helper.rb
```diff
@@ -0,0 +1,152 @@
...
+        list.concat(disk.fetch("partitions", []))
+      end
+    end
+
+    def root_part
+      partitions.detect {|p| p["mount"] == "/" }
+    end
+
+    def doable?
+      if root_part.nil?
```
lslezak: NP: use a guard clause here, it saves some indentation:
```ruby
return true unless root_part.nil?

log.info ...
```

src/lib/storage/shadowed_vol_helper.rb
```diff
@@ -0,0 +1,152 @@
...
+    # Copy of the root partition with all the shadowed subvolumes marked for
+    # deletion and with the previously deleted subvolumes restored if they are
+    # not shadowed anymore
+    #
+    # In order to make the recovery of previously deleted subvolumes possible,
+    # it stores the deleted volumes in a class variable. This list needs to be
+    # cleaned between subsequent calls to the expert partitioner.
+    # @see .cleanup
+    def root_partition
+      log.info " shadowed subvolumes"
```
lslezak: ??? I guess it's just a debug log which will be removed or improved...

src/lib/storage/shadowed_vol_helper.rb
```diff
@@ -0,0 +1,152 @@
...
+        elsif existing_subvol["delete"]
+          log.info "Rejecting subvol deletion"
+          existing_subvol["delete"] = false
+        end
+      end
+      clean_aborted_subvols
+    end
+
+    def aborted_subvols
+      @@aborted_subvols ||= []
```
lslezak: I use a singleton instance for keeping persistent data (https://github.com/yast/yast-registration/blob/master/src/lib/registration/storage.rb#L31), IMHO it's a bit more consistent with "normal" objects. But I'm not against class attributes, I simply do not know which one is better... :grey_question: 

But we should be consistent across YaST modules and use one pattern preferably... a topic for discussion?

test/fixtures/msdos-btrfs-shadowed-twice.ycp
```diff
@@ -0,0 +1,246 @@
+$[
```
lslezak: NP: Could we get rid of the YCP data format? I know it's a bit out of scope of this PR but we should finally kill the last `*.ycp` zombies...

BTW there is a [`ycp2yml.rb` conversion tool](https://github.com/yast/yast-devtools/blob/master/ytools/yast2/ycp2yml.rb)...

src/lib/storage/shadowed_vol_helper.rb
ancorgs: Actually I have the feeling we abuse guard clauses sometimes. For short cases like this, I personally find the `if..else` approach more readable. A matter of taste.

src/lib/storage/shadowed_vol_helper.rb
ancorgs: Yep, it definitely looks like a leftover.

src/lib/storage/shadowed_vol_helper.rb
ancorgs: I think you are right. The singleton pattern is more common in YaST codebase and, with the current shape of this class, there is no strong reason to not use singleton. I will change it

lslezak:
Thanks, LGTM now! :+1:

kobliha:
@ancorgs Great job! :+1: Thank you!

PR #196, opened 2016-02-24
Added YaML writer

shundhammer:
https://trello.com/c/avIbJtoB/303-2-libstorage-write-mockup-hardware-partitioning-tree-to-yaml-files

### DIFF
--- a/src/examples/fake_probing_demo.rb
+++ b/src/examples/fake_probing_demo.rb
@@ -10,5 +10,7 @@ sdx = ::Storage::Disk.create(devicegraph, "/dev/sdx")
 sdy = ::Storage::Disk.create(devicegraph, "/dev/sdy")
 sdz = ::Storage::Disk.create(devicegraph, "/dev/sdz")
 fake_probing.to_probed
+
+probed = Yast::Storage::StorageManager.instance.probed
 puts("Probed disks:")
-fake_probing.dump_disks(Yast::Storage::StorageManager.instance.probed)
+probed.all_disks.each { |disk| puts("  Found disk #{disk.name}") }
--- a/src/examples/yaml_writer_demo.rb
+++ b/src/examples/yaml_writer_demo.rb
@@ -4,24 +4,17 @@ $LOAD_PATH.unshift(File.expand_path('../../lib', __FILE__))
 
 require "storage/fake_probing.rb"
 require "storage/fake_device_factory.rb"
+require "storage/yaml_writer.rb"
 
-FILENAME = "fake-devicegraphs"
 input_file  = ARGV[0] || "fake-devicegraphs.yml"
+output_file = ARGV[1] || "/dev/stdout"
 
 fake_probing = Yast::Storage::FakeProbing.new
 devicegraph = fake_probing.devicegraph
 factory = Yast::Storage::FakeDeviceFactory.new(devicegraph)
 factory.load_yaml_file(input_file)
 
-fake_probing.to_probed
-puts("Disks:")
-probed = Yast::Storage::StorageManager.instance.probed
+yaml_writer = Yast::Storage::YamlWriter.new
+yaml_writer.write(devicegraph, output_file)
 
-# Write to graphviz format, convert to .png and display
-probed.write_graphviz("#{FILENAME}.gv")
-system("dot -Tpng <#{FILENAME}.gv >#{FILENAME}.png")
-system("display #{FILENAME}.png")
 
-# Clean up
-File.delete("#{FILENAME}.gv")
-File.delete("#{FILENAME}.png")
--- a/src/lib/storage/disk_size.rb
+++ b/src/lib/storage/disk_size.rb
@@ -204,7 +204,7 @@ module Yast
         unit_index = 0
         size = @size_k.to_f
 
-        while size > 1024.0 && unit_index < UNITS.size - 1
+        while size >= 1024.0 && unit_index < UNITS.size - 1
           size /= 1024.0
           unit_index += 1
         end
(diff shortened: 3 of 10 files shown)


src/examples/fake_probing_demo.rb
```diff
@@ -10,5 +10,7 @@
 sdy = ::Storage::Disk.create(devicegraph, "/dev/sdy")
 sdz = ::Storage::Disk.create(devicegraph, "/dev/sdz")
 fake_probing.to_probed
+
+probed = Yast::Storage::StorageManager.instance.probed
 puts("Probed disks:")
-fake_probing.dump_disks(Yast::Storage::StorageManager.instance.probed)
+probed.all_disks.each { |disk| puts("  Found disk #{disk.name}") }
```
ancorgs: Doesn't this prove that `FakeProbing` was not the right place to put `dump_disks`. :wink:

src/examples/yaml_writer_demo.rb
```diff
@@ -0,0 +1,20 @@
...
+
+fake_probing = Yast::Storage::FakeProbing.new
+devicegraph = fake_probing.devicegraph
+factory = Yast::Storage::FakeDeviceFactory.new(devicegraph)
+factory.load_yaml_file(input_file)
+
+yaml_writer = Yast::Storage::YamlWriter.new
+yaml_writer.write(devicegraph, output_file)
+
+
```
ancorgs: I prefer spec files over scripts in `/src/examples`. I know you don't like RSpec as documentation, but documentation tends to get outdated (nobody runs or revisit scripts at `/src/examples`) while tests are kept up-to-date by definition.

src/lib/storage/disk_size.rb
```diff
@@ -204,7 +204,7 @@ def to_human_readable
         unit_index = 0
         size = @size_k.to_f
 
-        while size > 1024.0 && unit_index < UNITS.size - 1
+        while size >= 1024.0 && unit_index < UNITS.size - 1
```
ancorgs: Is this the fix of a bug? If so, at some point (better sooner than later) we should add a test to avoid regression.

src/examples/fake_probing_demo.rb
shundhammer: This had been a debugging function to begin with. Now we can dump the entire tree to YAML wich is a lot more informative.

src/examples/yaml_writer_demo.rb
shundhammer: I prefer examples to be able to work with my code. This is not intended as a substitute for full-fledged test suites, but as a tool for working. It is not a good idea to deprive us of everything that makes the ever-growing complexity of our work environment worse all the time. We need something to invoke things directly to see what we are doing while we work on the code.

src/lib/storage/fake_device_factory.rb
```diff
@@ -224,6 +160,7 @@ def create_disk(parent, args)
       def create_partition_table(parent, args)
         log.info("#{__method__}( #{parent}, #{args} )")
         disk_name = parent
+        args = "msdos" if args.downcase == "ms-dos" # Allow different spelling
```
ancorgs: I fear this is now a little bit hidden here. I would look for a way to include this equivalence in `EnumMapping` in a more declarative way or would drop support for it. I don't like the idea of having some `if` here and there in addition to the mixin.

src/examples/yaml_writer_demo.rb
shundhammer: Plus, examples also serve the purpose in showing people how to use things. This won't work for very long if everything is always only available embedded in some deeply complex code that is not accessible from the outside (i.e. you can't call that code directly).

src/lib/storage/disk_size.rb
shundhammer: It's beautification: It gets you "1 TiB" instead of "1024 GiB" which is not exactly wrong, but I prefer the simpler way. BTW DiskSpace is one class that does have an extensive test suite, yet this was not caught. Yes, another test case could be added.

src/lib/storage/yaml_writer.rb
```diff
@@ -0,0 +1,141 @@
...
+            gap = partition.region.start - first_free_cyl
+            if gap > 0
+              partitions << yaml_free_slot(cyl_size*gap)
+            end
+
+            partitions << yaml_partition(partition)
+            first_free_cyl = partition.region.end + 1 unless partition.type == ::Storage::EXTENDED
+          end
+          content["partitions"] = partitions unless partitions.empty?
+        rescue RuntimeError => ex # FIXME: rescue ::Storage::Exception when SWIG bindings are fixed
```
ancorgs: We are starting to have this in too many places. Do we have a concrete plan to fix those SWIG bindings? We should push for it to be done as soon as possible.

src/examples/yaml_writer_demo.rb
ancorgs: I hate to start another discussion about APIs, but I have the feeling that we are going in the direction of having a `object_used_only_once = Whatever.new` every couple of lines. Something I fear can reveal wrong usage of object orientation and the single responsibility principle. And actually, this script is an perfect example of that because it contains, literally, a `Whatever.new` every couple of lines.

I believe that something like this feel way more natural for the average ruby developer (I'm not saying is perfect, though):
```ruby
factory = DevicegraphFactory.new
devicegraph = factory.create_from_file(input_file)
File.open(output_file, 'w') {|f| f.write devicegraph.to_yaml }
```
I did some prove of concept during the discussion of the previous pull request, and something like that can work without root permissions with no need to involve FakeProbing.

ancorgs:
Last philosophical comment (for the time being) :wink: I didn't expected to have the logic about the devicegraph hierarchy (disks that have partitions that have filesystems) in both the factory and the YAML writer.

I can perfectly live with that duplication for the time being, but looks like we didn't go far enough with our abstract and introspection effort.

PR #197, opened 2016-02-25
Cosmetic for one-trick ponies: Additional singleton methods

shundhammer:
This is for the most common use case of many of those classes: You create an instance and use it once to let it execute its main functionality.


wfeldt:
lgtm

kobliha:
Except that `All checks have failed`

ancorgs:
The "all checks have failed" is expected until we have a separate repository

PR #199, opened 2016-03-02
[Review] Request from 'schubi2' @ 'yast/yast-storage/review_160302_efi_boot_formating'

schubi2:
Please review the following changes:
  * 0bc6862 packaging
  * 2b7cb47 Efi boot: Select Efi Boot file system ID and allowing formating at the same time


src/include/partitioning/ep-dialogs.rb
```diff
@@ -48,6 +48,15 @@ def MiniWorkflowStepRole(data)
 
       role = data.value.fetch("role", :data)
 
+      efi_select = Empty()
+      if Partitions.EfiBoot && (Arch.i386||Arch.x86_64)
```
aschnell: ARM is also EFI.

package/yast2-storage.changes
```diff
@@ -17,6 +24,7 @@ Mon Mar  7 16:17:03 UTC 2016 - ancor@suse.com
 
 - Removed support for flexible partitioning proposal, a feature
   not used in SLE or openSUSE since years (fate#320111)
+>>>>>>> master
```
aschnell: Hmm.

aschnell:
Do I see it right that you can only create a EFI Boot Partition but not change an existing to an EFI Boot Partition? Is that clear to the usability experts?

src/include/partitioning/ep-dialogs.rb
```diff
@@ -441,6 +464,14 @@ def MiniWorkflowStepFormatMount(orig_data, creating: false)
       UI.ChangeWidget(Id(:do_mount_attachment), :Enabled, do_mount)
       UI.ChangeWidget(Id(:mount_point), :Value, mount)
 
+      if role == :efi_boot
```
schubi2: As far I see role will be set only for new created partitions (in MiniWorkflowStepRole). So while the update workflow these entries should not be disabled.

schubi2:
So this hits new generated partitions only. This has been approved by the UI experts.

aschnell:
OK. LGTM.

schubi2:
Thanks !

PR #201, opened 2016-03-07
Never use flexible_partitioning (fate#320111)

ancorgs:
Tested manually for normal installation and autoinstallation.


shundhammer:
LGTM

PR #203, opened 2016-03-11
Add PowerNV support

k0da:
Fixes boo#970582


k0da:
I there a way to tell travis to pick specific libstorage version?

k0da:
local rake osc:build succeedes here:
[  125s] 
[  125s] 		=== yast2-storage Summary ===
[  125s] 
[  125s] # of expected passes		35

lslezak:
I'd suggest ignoring the Travis failure for now, if the local osc build pasess then it's OK IMHO. I'll check why it fails at Travis but it will take some time as Travis debugging is difficult...

k0da:
ping

shundhammer:
LGTM

PR #206, opened 2016-04-10
Fix partitioning proposal for BTRFS on aarch64 (bsc#974280)

dirkmueller:


mvidner:
Looks Good To Me

package/yast2-storage.changes
```diff
@@ -1,4 +1,9 @@
 -------------------------------------------------------------------
+Sun Apr 10 18:27:50 UTC 2016 - dmueller@suse.com
+
+- raise min /boot size for BTRFS/aarch64 (bsc#974280)
```
imobach: Please, add the version number.

mvidner: I did so, in the merge commit.

mvidner:
Thank you, @dirkmueller !

src/modules/Partitions.rb
```diff
@@ -225,7 +225,11 @@ def BootSizeK
         }
         Ops.set(@boot_size_k, :proposed, 150 * 1024) if EfiBoot()
 
-        if Arch.ia64
+        if Arch.aarch64
```
k0da: POWER is 64k too

PR #209, opened 2016-04-13
Don't allow to create or change partition table on LDL DASDs  (bsc#958893)

gabi2:


gabi2:
I'm going to improve condition in Partitions.rb.

PR #211, opened 2016-05-05
/var/cache subvolume + one backport from SP1

jsrain:


ancorgs:
LGTM

kobliha:
@aschnell @ancorgs @shundhammer Considering this PR, there might be a good reason for writing such tests in a bit different way in the future, such as writing tests focused on a given functionality only (a real unit testing). What do you think? I hope it's done in a different way in new storage already.

PR #212, opened 2016-05-06
fixed error in porting bsc#957913 to master branch

jsrain:


jsrain:
This time tested with Peter Varkoly's control file

mvidner:
For reference: This is a fixup of #211  which was a port of #176.

LGTM

PR #216, opened 2016-07-18
- fixed proposed size of /boot/zipl on DASDs with GPT (bsc#988526)

aschnell:


src/modules/StorageProposal.rb
```diff
@@ -668,7 +668,7 @@ def try_add_boot(conf, disk, force)
         pb = {}
 	pb["mount"] = Partitions.BootMount
 	pb["size"] = Partitions.ProposedBootsize
-	if disk.fetch("label","")=="gpt" && !Partitions.EfiBoot
+	if disk.fetch("label","")=="gpt" && !Partitions.EfiBoot && !is_dasd?(disk.fetch("device", ""))
```
mvidner: Why does changes say "on DASDs with GPT" but this condition means "on nonDASDs with GPT (without efiboot)"?

aschnell: Because the code after the if overrides the boot size with a wrong value.

The old code was inserted to change the boot size to 1 MiB on non-EFI system on disks with GPT. The options are somewhere later changed to turn the partition into an bios-grub partition.

In the new bug we have a S/390 (non-EFI) with a FBA DASD that has a GPT. But we do not want the size for boot to be reduced to 1 MiB (it should remain at 200 MiB).

mvidner: >  change the boot size to 1 MiB on non-EFI system on disks with GPT

Yet I see `sz = 200*1024 if sz<200*1024` below which I interpret as "make the size at least 200*1024 KiB" and that confuses me.

mvidner:
The code still does not make sense to me but Arvin is able to defend it and the tests do make sense.
LGTM.

src/modules/StorageProposal.rb
aschnell: It also confuses me and I also think the code is wrong: I suppose ```sz``` should be ```cyl_size / 1024``` instead of ```- 1024```. But I don't want to change more than required.

aschnell: I looked at the code again and the ```size``` there is in B, not KiB. So it's at least 200 KiB, not 200 MiB.

That also means the ```- 1024``` should not be ```/ 1024```. Maybe it's still a typo that just doesn't hurt. Maybe it should be removed. Who knows.

PR #217, opened 2016-07-29
Don't ask twice about multipath and LUKS bsc#989770

ancorgs:
Goes together with https://github.com/yast/yast-installation/pull/410


jreidinger:
LGTM

PR #218, opened 2016-07-29
Don't oversize /boot in aarch64 with UEFI bsc#984874

ancorgs:
Tested manually. Seems to work.


jsrain:
Well, this is exactly what I would propose as patch.

PR #219, opened 2016-08-01
- fixed using complete multipath disk as LVM physical volume (bsc#984245)

aschnell:
Customer already confirmed that patch fixes the issue, see bugzilla.suse.com.


aschnell:
I have seen the jenkins failure but I doubt it is caused by the pull request.

wfeldt:
lgtm

PR #220, opened 2016-08-10
Fix for bsc#991252

ancorgs:
Fixes a bug when proposal suggests to turn a EPS into bios_grub

### DIFF
--- a/src/modules/Storage.rb
+++ b/src/modules/Storage.rb
@@ -6193,22 +6193,15 @@ module Yast
             id = Partitions.FsidBoot(dlabel)
             part["format"]=false
             part["mount"]=""
-            part["fstype"]=Partitions.FsIdToString(id)
             part["prep_install"]=true
-            if !part.fetch("create",false) &&
-               part.fetch("fsid",0) != id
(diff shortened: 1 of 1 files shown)


aschnell:
LGTM

aschnell:
Needs libstorage 2.26.6 for the testsuite.

ancorgs:
Test LGTM. Let's wait for OBS (libstorage) to re-run Travis.

testsuite/tests/bsc-991252.rb
```diff
@@ -0,0 +1,34 @@
+# encoding: utf-8
+
+# testedfiles: helper.rb
+
```
kobliha: I have just one comment here: `bsc-991252` is a great reference, but as the history shows, we often fix something which was actually not reported, because the root of the issue was something else. So, it would make sense to put some comment/information into the actual test that describes the scenario and test-case itself in human-readable form and explains what is the test about.

I like the way you create these tests for bugs, that's a good idea. I'm just looking into the future - how to benefit from that even more. Thanks!

PR #221, opened 2016-08-16
- Fixed message suggesting to add PReP partition (bsc#988783)

aschnell:
For https://trello.com/c/RZnByOJj/423-3-12-sp2-p1-bug-988783-sles12-sp2-beta4-configuring-a-ppc-prep-boot-partition-on-a-gpt-partition-table-fails-with-an-error.


sriemer:
Also looks good to me. Thanks!

PR #222, opened 2016-08-19
added /var/lib/machines to the list of subvolumes (bsc#992573)

jsrain:


aschnell:
Technical LGTM, conceptual of course broken.

jsrain:
@aschnell That's something we should fix for SP3

PR #223, opened 2016-08-22
SLE-11 SP4: Use stdin, not tmp files for passwords (bsc#986971)

shundhammer:
Backport to SLE-11 SP4.

This is very close (but not exactly the same) as the backport to SLE-12 GA. Prior to SLE-12 GA, libstorage was part of the yast2-storage package, thus the different package name here.

Also, in that version there were considerably less constants defined for all kinds of external tools; they were all over the place in the code. This is why the SLE-12 GA patch did not apply cleanly in Volume.cc and Volume.h.

SLE-12 GA patch for comparison:

https://github.com/openSUSE/libstorage/pull/163

_Please notice that this pull request is strictly about the patch being equivalent to the ones for SLE-12 GA and SLE-12 SP1. I tried to keep the patch as minimalistic as possible, so please let's avoid style discussions. The patch is tested and proven to work for SLE-12 GA and SP1._


libstorage/src/SystemCmd.cc
```diff
@@ -500,6 +551,34 @@ SystemCmd::checkOutput()
...
+    {
+        string::size_type count = 0;
+        string::size_type len   = _stdinText.size();
+        int result = 1;
+
+        while ( count < len && result > 0 )
+            result = fputc( _stdinText[ count++ ], _childStdin );
+
+        _stdinText.erase( 0, count );
+        // y2deb( count << " characters written; left over: \"" << _stdinText << "\"" );
```
lslezak: NP: if  you do not need this line then simply remove it, avoid dead code...

shundhammer: Well, in principle, yes; it's a matter of habit to keep useful debugging code (albeit commented out) in case it's useful in the future again for any debugging.

libstorage/src/Volume.cc
```diff
@@ -2186,17 +2199,18 @@ EncryptType Volume::detectEncryption()
 			break;
 		    }
 		bool excTime, excLines;
-		c.executeRestricted( cmd, 15, 500, excTime, excLines );
-		bool ok = c.retcode()==0 || (excTime && !excLines);
-		y2mil("ok:" << ok << " retcode:" << c.retcode() << " excTime:" << excTime <<
-		      " excLines:" << excLines);
+                SystemCmd fsckCmd;
+                fsckCmd.executeRestricted( cmd, 15, 500, excTime, excLines );
+                bool ok = fsckCmd.retcode()==0 || (excTime && !excLines);
```
lslezak: NP: It seems that adding `success()` method could avoid that repeating `retcode() == 0`...

shundhammer: Well, yes, there are many things that could be improved in that old code, but the approach here is to keep the diff minimalistic. This considerably improves the chances that the patch cleanly applies with increasingly older releases.

libstorage/src/Volume.cc
```diff
@@ -2390,30 +2384,35 @@ int Volume::doCryptsetup(bool readonly)
...
 		    if( ret==0 && mp=="swap" && crypt_pwd.empty() && !format )
-			cmd.execute("/sbin/mkswap " + quote(dmcrypt_dev));
+			SystemCmd mkswapCmd("/sbin/mkswap " + quote(dmcrypt_dev));
 		    }
 		}
 	    if( ret==0 && (!isTmpCryptMp(mp)||!crypt_pwd.empty()) )
 		{
-		string cmdline = getCryptsetupCmd( encryption, dmcrypt_dev, mp, fname, false );
-		if( !cmdline.empty() )
+		SystemCmd * cryptSetupCmd = createCryptSetupCmd( encryption, dmcrypt_dev, mp, crypt_pwd, false );
```
lslezak: This patterns repeats several times... But given this is just a backport I'm fine with that.

lslezak:
Adding a small <sub><sup>LGTM</sup></sub> as I'm not a C++ nor a storage expert. :wink:

PR #224, opened 2016-08-22
SLE-11 SP3: Use stdin, not tmp files for passwords (bsc#986971) 

shundhammer:
Backport to SLE-11 SP3.

The patch from SLE-11 SP4 applied cleanly here.

See also https://github.com/yast/yast-storage/pull/223


lslezak:
As this is basically the same patch as in #223 I'm giving my small <sub><sup>LGTM</sup></sub> here as well. :wink:

PR #225, opened 2016-08-26
Self update cleanup

imobachgs:
Don't check for installer restart as it's not needed anymore after moving self-update to the beginning of the installation.


lslezak:
LGTM

ancorgs:
I would also remove the `multipath_off?` method, that is not used anymore (if we merge https://github.com/yast/yast-installation/pull/429). Or maybe you want to merge that before removing the method, just to stay safe

imobachgs:
Good point. I'll keep the method by now to minimize the chance to break staging. We can remove it next week :)

PR #226, opened 2016-08-30
SLE-11 SP1: Use stdin, not tmp files for passwords (bsc#986971)

shundhammer:
Backport to SLE-11 SP1.

Since there were a lot of changes between SP1 and SP3, the SP3/SP4 patch did not apply at all, so this is very much a manual effort.

Still, it is quite straightforward to compare this PR manually to the PR against SP4:

https://github.com/yast/yast-storage/pull/223/files

I also created RPMs in the IBS with this version, built a DUD, integrated that DUD into the SLES-11-SP1 ISO and tested with that ISO: It installs without problem, and the created encrypted LVM can successfully be decrypted so that system boots without any problems.


libstorage/src/SystemCmd.cc
```diff
@@ -256,6 +288,11 @@ SystemCmd::doExecute( string Cmd )
 		    y2error( "dup2 stderr child failed errno=%d (%s)", errno,
 			     strerror(errno));
 		    }
+		if( close( sin[1] )<0 )
+		    {
+		    y2error( "close child failed errno=%d (%s)", errno,
```
ancorgs: NP (I wouldn't change already tested code for this): the original patch used "close (stdin) child failer errno", which is probably convenient for easily distinguish from the error closing sout below.

libstorage/src/Volume.cc
```diff
@@ -2007,19 +2006,19 @@ int Volume::doCryptsetup()
...
 	    ProcPart p;
 	    unsigned long long sz;
 	    if( p.getSize( Dm::dmDeviceName(minor), sz ))
 		setSize( sz );
 	    }
-	}
+        }
     else
-	{
+        {
```
ancorgs: This indeed fix the indentation according to the rest of the code. But I couldn't let pass the opportunity to say this: what an awful indentation! :wink:

ancorgs:
After a close inspection, I confirm this patch looks 100% equivalent to the SP4 one, so LGTM.

lslezak:
I have compared it to the https://github.com/yast/yast-storage/pull/223/files change and nothing was forgotten. So again, small <sub><sup>LGTM</sup></sub> from me :wink:

libstorage/src/Volume.cc
shundhammer: Yes, indeed it is. ;-(

The first thing I did when I started working on libstorage last year in April was to fix that indentation. We actually did have a mandatory coding style for C/C++ even back then, but certain people chose to ignore that completely.

libstorage/src/SystemCmd.cc
shundhammer: Right, but I wanted to keep it as similar to the surrounding code as possible -- and there it's similarly impossible to tell which close() went wrong - stdout or stderr? Now it's one more to guess from. But when that happens, something went badly wrong already, so it's just cosmetic which of those close() calls failed.

PR #229, opened 2016-09-14
Don't try to format a partition with file system type 'unknown' 

shundhammer:
Fix for https://bugzilla.suse.com/show_bug.cgi?id=996007


ancorgs:
LGTM.

It's more a workaround than a fix to the root evil (something that changed in libstorage that is leading to a stupid decision in the proposal) but it's fine... I just fear the root evil will manifest in other ways.

PR #234, opened 2016-10-27
Make subvolumes configurable in control.xml

shundhammer:
https://fate.suse.com/321737

https://trello.com/c/JdvwBp5G/457-5-microos-12-sp2-move-default-subvolumes-to-control-file

### DIFF
--- a/src/lib/storage/subvol.rb
+++ b/src/lib/storage/subvol.rb
@@ -33,6 +33,32 @@ module Yast
 
       attr_accessor :path, :copy_on_write, :archs
 
+      COW_SUBVOL_PATHS = [
+        "home",
+        "opt",
+        "srv",
+        "tmp",
+        "usr/local",
+        "var/cache",
+        "var/crash",
+        "var/lib/machines",
+        "var/lib/mailman",
+        "var/lib/named",
+        "var/log",
+        "var/opt",
+        "var/spool",
+        "var/tmp"
+      ]
+
+      # No Copy On Write for SQL databases and libvirt virtual disks to
+      # minimize performance impact
+      NO_COW_SUBVOL_PATHS = [
+        "var/lib/libvirt/images",
+        "var/lib/mariadb",
+        "var/lib/mysql",
+        "var/lib/pgsql"
+      ]
+
       def initialize(path, copy_on_write: true, archs: nil)
         @path = path
         @copy_on_write = copy_on_write
@@ -42,7 +68,8 @@ module Yast
       def to_s
         text = "Subvol #{@path}"
         text += " (NoCOW)" unless @copy_on_write
-        text += archs.nil? ? " (archs: all)" : " (archs: #{@archs})"
+        text += " (archs: #{@archs})" if arch_specific?
+        text
       end
 
       def arch_specific?
@@ -55,6 +82,12 @@ module Yast
 
       def no_cow?
         !@copy_on_write
+      end
+
+      # Comparison operator for sorting
+      #
+      def <=>(other)
+        path <=> other.path
       end
 
       # Check if this subvolume should be used for the current architecture.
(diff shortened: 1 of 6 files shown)


src/lib/storage/subvol.rb
```diff
@@ -0,0 +1,171 @@
+# encoding: utf-8
+
+# Copyright (c) [2012-2015] Novell, Inc.
```
imobachgs: Maybe you would like to update the copyright?

src/lib/storage/subvol.rb
```diff
@@ -0,0 +1,171 @@
...
+      def <=>(other)
+        path <=> other.path
+      end
+
+      # Check if this subvolume should be used for the current architecture.
+      # A subvolume is used if its archs contain the current arch.
+      # It is not used if its archs contain the current arch negated
+      # (e.g. "!ppc").
+      #
+      # @return bool
```
imobachgs: The format should be something like `@return [Boolean] description...

src/lib/storage/subvol.rb
```diff
@@ -0,0 +1,171 @@
...
+          if block_given?
+            match = block.call(arch)
+          else
+            match = arch == target_arch
+          end
+          if match && negate
+            log.info("Not using #{self} for explicitly excluded arch #{arch}")
+            return false
+          end
+          use_subvol ||= match
```
imobachgs: Maybe you would like to use `archs.any?` instead of `each` so you don't need to take care of use_subvol. Moreover, if you find the architecture, the loop will end.

test/subvol_test.rb
```diff
@@ -0,0 +1,191 @@
...
+describe Yast::StorageClass::Subvol do
+
+  context "#new" do
+    let(:current_arch) { Yast::Arch.arch_short }
+
+    describe "Simple subvol with defaults" do
+      subject { Yast::StorageClass::Subvol.new("var/spool") }
+
+      it "has the correct path" do
+        expect(subject.path).to be == "var/spool"
```
imobachgs: `.to eq "var/spool"` is shorter :)

test/subvol_test.rb
```diff
@@ -0,0 +1,191 @@
...
+    describe "arch-specific subvol for everything except the current arch" do
+      subject { Yast::StorageClass::Subvol.new("boot/grub2/fake-arch", archs: ["!#{current_arch}"]) }
+
+      it "does not match the current arch" do
+        expect(subject.current_arch?).to be false
+      end
+    end
+  end
+
+  context "#create_from_xml" do
```
imobachgs: A class-method should be described as `.create_from_xml`.

test/subvol_test.rb
```diff
@@ -0,0 +1,191 @@
+#!/usr/bin/env rspec
+
+require_relative "spec_helper"
+require "storage/subvol"
+require "pp"
```
imobachgs: You don't need it, right?

test/subvol_test.rb
```diff
@@ -0,0 +1,191 @@
...
+Yast.import "Arch"
+
+
+describe Yast::StorageClass::Subvol do
+
+  context "#new" do
```
imobachgs: I would prefer a different approach focused in each method:

```ruby
describe "#no_cow?" do
  subject { Subvol.new("/var/lib") }

  context "by default" do
    it "returns true" do
      expect(subject.no_cow?).to eq(true)
    end
  end

  context "when is not a nocow volume" do
    subject { Subvol.new("/var/lib/pgsql", noco) }

    it "returns false" do
      expect(subject.no_cow?).to eq(false)
    end
  end

  # ...
end
```

But it's fine for me as it.

test/subvol_test.rb
shundhammer: No, "pp" is not needed here.

test/subvol_test.rb
shundhammer: Shorter maybe, but not necessarily better readable IMHO.

test/subvol_test.rb
shundhammer: ok

src/lib/storage/subvol.rb
shundhammer: Ah, right.

PR #235, opened 2016-10-31
Set default subvolume name

imobachgs:
This PR improves the support to redefine/detect the used default_subvol. For the time being, only `@` and `""` are supported (checked with @aschnell).


imobachgs:
Thanks!

PR #238, opened 2016-11-03
Add subvol.rb to Makefile.am

imobachgs:
Add the missing `lib/storage/subvol.rb` file. I'm not bumping the version number because version 3.1.103 has not been submitted yet.


shundhammer:
LGTM

PR #242, opened 2016-11-11
Revert FATE#320342 and FATE#317775

imobachgs:
I've reverted changes for FATE#320342 and FATE#317775 as they won't be present in SLE 12 SP2 and, moreover, it introduces a bug (that to be fixed needs to update yast-autoinstallation).

A `git revert -m 1 a29db9b` could do the trick (you'll need to un-revert the changelog/version changes) but I've preferred to revert commit by commit to make revision easier (I'll squash all of them when merging).


jreidinger:
but then we have to somehow merge it from CASP branch not? and does not this revert break that merge and we need do double revert?

imobachgs:
@jreidinger I don't expect this change to be merged from CASP (and even I don't want it). This feature should be only in CASP and SP3 as it's related to the installation.

On the other hand, I could just create another PR that includes those changes using simply a patch and avoiding the revert command.

jreidinger:
ok, so LGTM

imobachgs:
Thanks!

PR #243, opened 2016-11-17
Don't show [Edit Proposal Settings] button if disabled in control.xml (Fate#321736)

shundhammer:
- Don't show [Edit Proposal Settings] button if disabled in control.xml

- Added warning popup for expert partitioner if configured in control.xml


ancorgs:
LGTM

PR #246, opened 2016-11-30
Avoid confusion in small partition warning by displaying also rounded input value (bsc#910791)

dwaas:
See https://github.com/yast/yast-storage/pull/231

Another PR was opened due to a git push -f.

### DIFF
--- a/src/include/partitioning/custom_part_lib.rb
+++ b/src/include/partitioning/custom_part_lib.rb
@@ -293,13 +293,17 @@ module Yast
 
       if Ops.less_than(size_k, min_size_k)
         # warning message, %1 is replaced by fs name (e.g. Ext3)
-        # %2 is prelaced by a size (e.g. 10 MB)
+        # %2 is replaced by the user entered size (e.g. 20GB)
+        # %3 is replaced by a size (e.g. 10 MB)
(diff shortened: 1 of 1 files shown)


src/include/partitioning/custom_part_lib.rb
```diff
@@ -291,15 +291,19 @@ def check_ok_fssize(size_k, volume)
...
+      if size_k < min_size_k
         # warning message, %1 is replaced by fs name (e.g. Ext3)
-        # %2 is prelaced by a size (e.g. 10 MB)
+        # %2 is replaced by the user entered size (e.g. 20GB)
+        # %3 is replaced by a size (e.g. 10 MB)
         Popup.Warning(
           Builtins.sformat(
             _(
-              "Your partition is too small to use %1.\nThe minimum size for this file system is %2.\n"
+              "Your partition is too small to use %1.\n" +
```
imobachgs: Ruby allows you to use a different approach to avoid string concatenation (which is expensive):

```ruby
Builtins.sformat(
  _("Your partition is too small to use %1.\n" \
     "The size you entered (after rounding up) is %2.\n" \
     "The minimum size for this file system is %3.\n")
)
```

PR #249, opened 2017-01-10
Backport stuff to unify versions and fix bug expert partitioner bug

ancorgs:
This pull request includes two backports to fix the versions number mess (see my mail at yast-devel) and also a fix for https://bugzilla.suse.com/show_bug.cgi?id=1008740.

Tests for the fix included. Also tested manually

PR #250, opened 2017-01-10
Merge pull request #249 into CASP and fix versioning

ancorgs:
This includes a revert of the original revert made in SP2, so now git can track correctly the relation between both branches (SP2 and CASP).


shundhammer:
LGTM

PR #256, opened 2017-01-25
Expert partitioner less restrictive with encryption

shundhammer:
https://trello.com/c/4IojS1Y5/490-2-12-sp3-fate-320215-allow-full-lvm-encryption-with-msdos-mbr-without-extra-boot

The FATE request was really about not forcing the user to create a separate /boot partition when using encryption: Grub2 learned how to handle booting from encrypted partitions, from encrypted LVM, even from encrypted RAIDs.

The expert partitioner never enforced that, but it wouldn't let you use a "system mount point" (/, /usr/, /var, /opt (???)) when selecting encryption for a partition. This restriction is now lifted.

Along with it, more restrictions are now lifted: It wouldn't allow a Btrfs partition or LVM logical volume to be encrypted (why??), and it wouldn't allow encryption on an LVM physical volume (which doesn't make any sense since this is exactly what the partitioning proposal does) or encryption on a partition used by a RAID. This is now possible, but the user still has to know what he is doing.

Caveat: The user has to enter the encryption password twice in some scenarios, once for Grub2 to load kernel and initrd, and once later when systemd starts mounting filesystems. This is a limitation that was accepted by the Fate stakeholders (and it would have to be fixed by the Grub or systemd people anyway).

Caveat 2: The expert partitioner tries to be smart - in some cases too smart. When wildly clicking around all the options on the "Add partition" dialog, there may be situations when some widgets are enabled or disabled when they really shouldn't. In that case, the user might need to go back to the previous step in that sub-workflow.

### DIFF
--- a/src/include/partitioning/ep-dialogs.rb
+++ b/src/include/partitioning/ep-dialogs.rb
@@ -208,7 +208,6 @@ module Yast
           :crypt_fs,
           :Enabled,
           Ops.get_boolean(fs_data, :crypt, true) &&
-            Ops.get_symbol(data, "used_fs", :unknown) != :btrfs &&
             !Ops.get_boolean(data, "pool", false)
         )
 
@@ -445,7 +444,7 @@ module Yast
       #not there in RAID/LVM/loop configuration (#483789)
       ChangeWidgetIfExists(:do_not_format_attachment, :Enabled, !do_format)
 
-      _EnableDisableFsOpts.call(used_fs) if do_format
+      _EnableDisableFsOpts.call(used_fs)
 
       #not there on s390s
       ChangeWidgetIfExists(:crypt_fs, :Value, crypt_fs)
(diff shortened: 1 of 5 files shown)


src/include/partitioning/ep-dialogs.rb
```diff
@@ -445,7 +444,7 @@ def MiniWorkflowStepFormatMount(orig_data, creating: false)
       #not there in RAID/LVM/loop configuration (#483789)
       ChangeWidgetIfExists(:do_not_format_attachment, :Enabled, !do_format)
 
-      _EnableDisableFsOpts.call(used_fs) if do_format
```
jreidinger: rest looks ok, I just do not see reason for this change?

shundhammer: That function is a misnomer: It doesn't only enable or disable the "filesystem options" button, but also the "encrypt" check box. And this needs to be called in **every** case, not only if the user wants to format a partition. Otherwise that check box never gets enabled for partition type LVM (LVM physical volume) or RAID.

That enabling/disabling code in that dialog is a big mess anyway, but now it becomes a little bit more usable. There is no reason not to check enabled/disabled state after a user action, even more so since in that dialog everything depends on everything else.

PR #261, opened 2017-03-13
block installation unless snapshots are enabled (bsc#1019652)

wfeldt:


wfeldt:
replaced by https://github.com/yast/yast-storage/pull/262

PR #263, opened 2017-03-14
block installation unless snapshots are enabled (bsc#1019652)

wfeldt:


package/yast2-storage.changes
```diff
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Mar 14 10:24:01 CET 2017 - snwint@suse.de
+
+- block installation unless snapshots are enabled (bsc#1019652)
```
ancorgs: This message will be merged soon to all the other branches, we should make clear that in only applies if simple_mode is used (CaaSP).

PR #265, opened 2017-03-20
Removed calls to /sbin/insserv

mchf:
bnc#1026027


ancorgs:
I guess this applies to SLE-12-SP2 as well. Wouldn't be a good idea to fix it there first and then merge to CASP and master instead of leaving SP2 out? This change does not seem to be related to the other only-in-CASP-so-far stuff, so I see no reason to postpone the integration in all the branches.

PR #266, opened 2017-03-20
Merged SLE-12-SP2

mchf:


mchf:
For reference https://lists.opensuse.org/yast-devel/2017-02/msg00006.html

PR #272, opened 2017-05-23
Fix Btrfs default subvolume name

imobachgs:
Fixes [bsc#1040154](https://bugzilla.novell.com/show_bug.cgi?id=1040154).


src/modules/Storage.rb
```diff
@@ -354,6 +354,9 @@ def InitLibstorage(readonly)
       @conts = getContainers
       log.info("InitLibstorage conts:#{@conts}")
 
+      # Initializes default subvolume name (bsc#1040154)
+      FileSystems.read_default_subvol_from_target if Mode.normal || Mode.config
```
lslezak: Um, and what about the "firstboot" mode? How it should behave in that mode?

(I'm not sure how much the storage module would make sense in the firstboot mode, but anyway...)

test/storage_test.rb
```diff
@@ -13,34 +13,30 @@
...
+    around do |example|
+      old_stage = Yast::Stage.stage
+      Yast::Stage.Set(stage)
+      example.run
+      Yast::Stage.Set(old_stage)
+    end
 
-    context "when running on normal mode" do
-      let(:mode) { "normal" }
+    context "when running on initial stage" do
```
lslezak: NP: I think the correct wording is `in the initial stage`.

PR #277, opened 2017-06-16
Fix default subvolume name detection

imobachgs:
This PR fixes several issues: [bsc#1044434](https://bugzilla.suse.com/show_bug.cgi?id=1044434), [bsc#1044250](https://bugzilla.suse.com/show_bug.cgi?id=1044250) and [bsc#988700](https://bugzilla.suse.com/show_bug.cgi?id=988700).

### DIFF
--- a/package/yast2-storage.spec
+++ b/package/yast2-storage.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-storage
-Version:        3.2.15
+Version:        3.2.16
 Release:        0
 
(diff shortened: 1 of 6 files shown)


package/yast2-storage.spec
```diff
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-storage
-Version:        3.2.15
+Version:        3.2.16
```
jreidinger: will it go to SP3 or SLE15?

imobachgs: SP· and SLE15

PR #283, opened 2017-07-20
Bsc 1049108 sle12 sp3

aschnell:
For https://trello.com/c/sBRQZEzW/670-3-caasp-p1-bnc-1049108-installer-internal-error-in-storageproposalrb.

Note: Due to the block of https://trello.com/c/zCKSIBHK/633-1-merge-caasp-hack-for-separate-data-partition-to-sle-12-sp3-branch that code is not merged to SLE12-SP3 along to way.

### DIFF
--- a/package/yast2-storage.changes
+++ b/package/yast2-storage.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Jul 20 10:45:18 CEST 2017 - aschnell@suse.com
+
+- fixed check whether snapshots are proposed (bsc#1049108)
+- 3.2.17
+
+-------------------------------------------------------------------
 Fri Jun 16 08:07:04 UTC 2017 - igonzalezsosa@suse.com
 
 - Fix Btrfs default subvolume name detection (bsc#1044434 and
--- a/package/yast2-storage.spec
+++ b/package/yast2-storage.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-storage
-Version:        3.2.16
+Version:        3.2.17
 Release:        0
 
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
--- a/src/modules/StorageProposal.rb
+++ b/src/modules/StorageProposal.rb
@@ -6566,7 +6566,7 @@ module Yast
       ret = false
       if GetProposalSnapshots()
         prop_target_map.each do |device, container|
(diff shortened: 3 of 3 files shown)


jreidinger:
missing commit is https://github.com/yast/yast-storage/pull/283/commits/faab8c4ddecaa44799f4e8e16c8856d97e3e4c97

aschnell:
Yes, the commit from Stefan Hundhammer is missing on purpose as explained in the initial comment.

jreidinger:
@aschnell I am just not sure if this is good way to do it. Correct way would be to revert properly HuHa patches, so when decided to do it, we simply can revert that reverts. Now it looks like merged, but cannot be reverted to include it.

aschnell:
So what now? Just stop and continue with the PBI in a few weeks?

jreidinger:
@aschnell well, if it is for caasp only as jsrain mention, then I suggest to do it at once. So postpone merge to SP3 and merge it to master only like HuHa did.

jreidinger:
problem with this merge is that it pretend that it merge all caasp fixes into SP3, which is not true.

aschnell:
Suppose the fix would be need in SP3 now, how would that be handled?

jreidinger:
@aschnell if it is that case, then do it as I suggested. Do merge, but keep HuHas changes, then add additiona revert commits which reasoning why it have to be reverted now.

so something like
```
git merge
# merge conflict, but do not trash HuHas changes
git revert <huha's commits>
```
this will allow easy future revert of your revert. and it is also clear why it is not yet in SP3.

jreidinger:
and as a bonus mention it in HuHa's PBI so he knows that only revert of revert is needed.

aschnell:
Sorry, that looks insane to me. The process is broken.

jreidinger:
@aschnell well, in general I found bad decision to have change in SP2 and CaaSP that is not in SP3, that is part which broke process. General rule also forced by PM and release managers is to not lost any changes over time. And this is exactly what happen here. We can of course invite to review someone else to review if he have better idea or if he see this change as OK like @ancorgs @imobachgs or @mvidner

teclator:
@aschnell @jreidinger the bug was closed submitting the changes directly to CAASP, but what is the state of the PR? still valid? Seems that it was already merged by #284 so will close it but feel free to reopen if you consider that something regarding to the discussion should be addressed or is relevant

PR #284, opened 2017-08-08
Merged SLE-12-SP2-CASP branch to SLE-12-SP3 branch

shundhammer:
https://trello.com/c/zCKSIBHK/633-1-merge-caasp-hack-for-separate-data-partition-to-sle-12-sp3-branch

# Test Scenarios:

part-ext4-sep-home

Partition-based with ext4 root and separate /home

------------------------------------

part-ext4-no-home

Partition-based with ext4 root and no separate /home

------------------------------------

part-btrfs-no-home

Partition-based with btrfs and no separate /home. Notice the home subvolume.

------------------------------------

part-btrfs-separate-home

LVM with btrfs and separate /home. Notice **no** home subvolume.

------------------------------------

lvm-btrfs-no-home

LVM with btrfs and no separate /home. Notice the home subvolume.

------------------------------------

lvm-btrfs-separate-home

LVM with btrfs and separate /home. Notice no home subvolume.

### DIFF
--- a/package/yast2-storage.spec
+++ b/package/yast2-storage.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-storage
-Version:        3.2.16
+Version:        3.2.17
 Release:        0
 
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
--- a/package/yast2-storage.spec
+++ b/package/yast2-storage.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-storage
-Version:        3.2.17
+Version:        3.2.16.1
 Release:        0
 
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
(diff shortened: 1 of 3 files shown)


package/yast2-storage.spec
```diff
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-storage
-Version:        3.2.16
+Version:        3.2.17
```
gilsonsouza: Should it not be 3.2.16.1 since we don't have the fix for bsc#1049108 in master, which is already 3.2.17?

shundhammer: Right, those clashing versions are a very valid point.

But probably we should have started numbering with 3.3.x for post-SP3.

shundhammer:
For now using version 3.2.16.1 to prevent clashing version numbers between the SP3 branch and master.

Once we come to a conclusion what version number scheme to use for SLE-15, we can return to three-digit version numbers in this branch, too.

PR #289, opened 2017-08-22
Casp merge

teclator:


package/yast2-storage.changes
```diff
@@ -1,30 +1,73 @@
...
-- 3.1.108 
+- 3.1.108.1 
 
 -------------------------------------------------------------------
 Mon Mar 20 09:12:25 UTC 2017 - mfilka@suse.com
 
 - bnc#1026027
   - dropped calls to /sbin/insserv
-- 3.1.107
+- 3.1.107.1
```
shundhammer: Not sure about all those version numbers in the change log. AFAICS, those versions never existed in the target branch (the branch we are merging to). I understood our policy towards this that we simply omit the version numbers from the change log entry in case of such a case; we just need to make sure that the timestamps are in the correct sequence (otherwise some check script will complain, and Travis won't build and OBS won't accept the submit request).

shundhammer:
Also, I suggest to bump the version in the .spec file and add another change log entry about this merge.

package/yast2-storage.changes
teclator: Basically we are using the CASP changelog, and it is in order, Travis is not complaining about it. Of course if we do a maintenance update for SP2 then we will change Sp2 changelog history, but the same will happen with CASP if we uses SP2 ones.

teclator:
It seems that yast/yast-schema#29 needs this version of yast2-storage so make sense to bump it as well yast/yast-autoinstallation#322

package/yast2-storage.changes
shundhammer: No, the order is okay. But since this is the SP2 branch, do we really want those CASP 1.0 only version numbers in this change log? I don't think anybody will ever remember to remove them if there is a maintenance update for SP2.

teclator: No one will remove them, but since now it will be the CASP and SP2 branch for maintenance updates so the changelog will be wrong for one of them being strict with the versions released

shundhammer:
Code-wise it's LGTM.

shundhammer:
[16:37:29] ‎<‎lslezak‎>‎ HuHa: there is no correct way IMHO,  SLES and CASP used different versions if we merge them together then either SLES will see non-existing versions or CASP :-(

teclator:
@shundhammer thnx for review!

PR #292, opened 2017-09-28
fixed missing changes

schubi2:


teclator:
@schubi2 the PR is probably out of date, and the version in the changelog also conflict with the entry above it. 

We should or remove the version in the changelog or close it (closed it by now)

PR #293, opened 2017-11-23
Replace references to /var/adm/fillup-templates with new %_fillupdir …

sysrich:
…macro (boo#1069468)

### DIFF
--- a/package/yast2-storage.spec
+++ b/package/yast2-storage.spec
@@ -16,6 +16,11 @@
 #
 
 
+#Compat macro for new _fillupdir macro introduced in Nov 2017
+%if ! %{defined _fillupdir}
+  %define _fillupdir /var/adm/fillup-templates
+%endif
+
 Name:           yast2-storage
 Version:        3.2.18
 Release:        0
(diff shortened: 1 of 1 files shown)


sysrich:
Ping - this PR is going to be required for any future yast-storage submission to be acceptable for SLE 15, Leap 15, or Tumbleweed, as /var/adm/fillup-templates is now a blacklisted path

ancorgs:
I just restart travis to see if we can merge this. Anyways, if the targets are SLE 15, Leap 15 and TW this could be less than urgent. SLE15 and Leap15 don't include yast-storage anymore (replaced by yast-storage-ng). And the plan is to also drop yast-storage from TW in a matter of days, weeks at most.

imobachgs:
I've built this package manually and it works just fine. So I guess we could skip Travis and leave Jenkins to stop it if something goes wrong.

ancorgs:
@imobachgs Makes sense. I already suspected the problem (the one visible in travis, at least) was our docker image being based on libstorage-ng. I guess it will build fine if the system running the tests is based on libstorage instead of libstorage-nb. So let's move forward.

PR #298, opened 2018-06-20
- mask systemd mount and swap units (bsc#1073633)

aschnell:
For https://trello.com/c/kkmPuoxR/330-3-1073633-l3-system-auto-mounts-ext4-partition-on-xen-pv-device-after-fs-resize-causes-yast-error.


joseivanlopez (approved):
LGTM

PR #300, opened 2018-08-02
Allow go back

teclator:
- https://bugzilla.suse.com/show_bug.cgi?id=1103408

I found that we disallowed to continue on this [PR](https://github.com/yast/yast-storage/pull/202), and probably the else should be a elsif condition

https://github.com/yast/yast-storage/pull/202/files#diff-82868f79ae37630ff5beeec65a4b6fa7R280

### DIFF
--- a/package/yast2-storage.changes
+++ b/package/yast2-storage.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Aug  2 18:14:46 UTC 2018 - knut.anderssen@suse.com
+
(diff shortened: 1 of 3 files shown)


aschnell:
The original code looks like a statement with no effect which some tool (e.g. rubocop) should recognise.

aschnell (approved):
LGTM

teclator:
Yep, but rubocop was not enabled in this project / repository so that was the reason it was not catched.

PR #301, opened 2018-08-23
Consistent Home Limit Check

shundhammer:
### Trello

https://trello.com/c/CRZ147Oh/186-5-sles12-sp4-p3-1085169-build-0230-openqa-test-fails-in-partitioninglvm-fails-to-propose-home-partition-put-proposed-itself

### Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1085169

# Problem

yast-storage-old checked the limit when to propose a separate home partition differently between

- partition-based proposal
- LVM-based proposal (both plain and encrypted)

In one case, it took the other partitions that needed to be created into account (swap, /boot, /efi-boot), in the other it didn't. This is inconsistent.

# Ideal Solution

Make the proposal ~~great again~~ consistent in all cases. Always take the other partitions into account.

# Pragmatic Solution

Take the swap partition into account and just ignore the boot partition(s). This is what this PR does.

While it would be nice to be exact, the code adding a boot partition is all over the place in StorageProposal.rb and its 117 functions in its 6722 lines of code. That old code can only handle one single boot partition; it's either `/boot` or PreP or an EFI boot partition or whatever; if it's anything else than `/boot`, it is modified for its true purpose as a final step. This does not make the code any clearer.

Anyway, since a boot partition is only a few MiB anyway, this doesn't change the overall calculation that much; this is about the same order of magnitude than overhead due to LVM management which in the other case would also cause a slight difference. I think we can live with that.

Trying not to overengineer this.

### DIFF
--- a/package/yast2-storage.changes
+++ b/package/yast2-storage.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Mon Aug 27 16:19:54 UTC 2018 - shundhammer@suse.com
+
+- Consistent home limit check (bsc#1085169)
+- 3.2.17
+
+-------------------------------------------------------------------
 Thu Aug  2 18:14:46 UTC 2018 - knut.anderssen@suse.com
 
 - Allowed to go back or cancel in the inst_disk_proposal client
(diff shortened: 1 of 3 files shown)


joseivanlopez (approved):
LGTM

PR #302, opened 2018-08-28
adjust target map to never contain the wrong prep type (bsc#1098618)

wfeldt:


shundhammer (approved):
It's not very Ruby'ish, but in that it fits the rest of that autoconverted code. :smiley: 

Once Travis is happy, LGTM.

PR #304, opened 2018-09-28
really update prep type in target map (bsc#1098618)

wfeldt:


jreidinger:
I expect that test for this change is too hard, right? Is it tested manually or by customer?

wfeldt:
Yes; tested manually and hopefully by customer.

jreidinger (approved):

PR #305, opened 2019-04-01
Fix "Not enough space available to propose separate /home."

mvidner:
a was-L3, applying the PTF that helped the customer, plus an improvement of the error message

https://bugzilla.suse.com/show_bug.cgi?id=1113714

### DIFF
--- a/src/clients/inst_target_part.rb
+++ b/src/clients/inst_target_part.rb
@@ -446,8 +446,9 @@ module Yast
                   Ops.get_string(p, "mount", "") == StorageProposal.GetHomePath()
                 end) == 0
               @ok = false
-              reason = _(
-                "Not enough space available to propose separate /home."
+              reason = Builtins.sformat(
+                _("Not enough space available to propose separate %1."),
+                StorageProposal.GetHomePath()
               )
               Popup.Error(reason)
             end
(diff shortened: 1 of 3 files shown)


shundhammer (approved):
LGTM

src/clients/inst_target_part.rb
```diff
@@ -443,11 +443,12 @@ def main
             Builtins.y2milestone("proposed partitions:%1", @pl)
             if StorageProposal.GetProposalHome &&
                 Builtins.size(Builtins.filter(@pl) do |p|
-                  Ops.get_string(p, "mount", "") == "/home"
+                  Ops.get_string(p, "mount", "") == StorageProposal.GetHomePath()
                 end) == 0
               @ok = false
-              reason = _(
-                "Not enough space available to propose separate /home."
+              reason = Builtins.sformat(
```
ancorgs, line 449: Why to use `Builtinst.sformat` (with its own placeholder format) instead of the more standard Ruby's `format` (which uses standard strformat placeholders).

mvidner: I think I wanted to be consistent with the rest of the translatable strings:
```console
$ grep -c '%1' storage.pot
295
$ grep -c '%s' storage.pot
1
```

PR #295, opened 2017-12-07
Escape %{VERSION} in the spec file

imobachgs:
* It fixes bsc#1071724.

Travis is expected to fail (see [this comment](https://github.com/yast/yast-storage/pull/293#issuecomment-348164397)).

PR #306, opened 2019-06-03
Fix "Not enough space available to propose separate /home." (SP4)

mvidner:
A merge of #305.

A was-L3, applying the PTF that helped the customer, plus an improvement of the error message

https://bugzilla.suse.com/show_bug.cgi?id=1113714

### DIFF
--- a/package/yast2-storage.changes
+++ b/package/yast2-storage.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon Apr  1 13:20:46 UTC 2019 - mvidner@suse.com
+
+- Fix "Not enough space available to propose separate /home."
+  (bsc#1113714)
+- 3.2.16.4
+
+-------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


shundhammer (approved):
LGTM

package/yast2-storage.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Mon Apr  1 13:20:46 UTC 2019 - mvidner@suse.com
+
+- Fix "Not enough space available to propose separate /home."
+  (bsc#1113714)
+- 3.2.20.1
```
ancorgs, line 6: Why adding a new digit? This is not diverging from any more modern branch, isn't it?

ancorgs: Replied in IRC: is because an already existing divergence between this branch and `master_old` :-(

mvidner: Because even 3.2.20 was colliding with #295 ... which was the last existing maser version, so 3.2.21 would actually not collide. But version numbers are cheap.

PR #310, opened 2022-04-06
Fix LVM VG resize

joseivanlopez:
## Problem

In some cases, the target map contains a LVM VG that uses udev names instead of kernel names for its LVM PVs. When this happens, the current PVs are not found in the list of PVs after resizing. This implies that YaST will accidentaly try to remove and add such PVs.

* https://bugzilla.suse.com/show_bug.cgi?id=1197208
* https://trello.com/c/O3JvxmSG/2909-bignum-too-big-to-convert-into-long-long-rangeerror

## Solution

Always use kernel names when looking for the PVs after resizing a VG.

## Testing

- Added new unit tests
- Tested manually by customer

I was not able to reproduce an scenario in which the PVs are pointed by udev names. A DUD was provided to the customer to check this fix.

### DIFF
--- a/package/yast2-storage.changes
+++ b/package/yast2-storage.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Apr  6 15:16:08 UTC 2022 - José Iván López González <jlopez@suse.com>
+
+- Partitioner: PVs are not wrongly removed when resizing a VG
+  (bsc#1197208).
+- 3.2.23
+
+-------------------------------------------------------------------
 Fri Sep 27 10:40:48 CEST 2019 - aschnell@suse.com
 
 - reuse at most 20 swap devices (plus potentially more handled
--- a/src/include/partitioning/ep-lvm-lib.rb
+++ b/src/include/partitioning/ep-lvm-lib.rb
@@ -100,6 +100,53 @@ module Yast
       nil
     end
 
+    # Finds the device in the given target map and returns its kernel name
+    #
+    # In case the device is not found, the given device name is returned
+    #
+    # @param target_map [Hash]
+    # @param device_name [String]
+    #
+    # @return [String]
+    def device_kernel_name(target_map, device_name)
+      device = find_device(target_map, device_name)
+
+      return device_name unless device
+
+      device["device"]
+    end
+
+    # Looks for the a device with the given name in the target map
+    #
+    # @param target_map [Hash]
+    # @param device_name [String]
+    #
+    # @return [Hash]
+    def find_device(target_map, device_name)
+      devices = []
+
+      target_map.values.each do |device|
+        devices << device
+        devices += (device["partitions"] || [])
+      end
+
+      devices.find { |d| match_any_name?(d, device_name) }
+    end
+
+    # Checks whether any of the device names matches with the given name
+    #
+    # @param device [Hash]
+    # @param name [String]
+    #
+    # @return [Boolean]
+    def match_any_name?(device, name)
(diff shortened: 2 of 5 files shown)


package/yast2-storage.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Wed Apr  6 15:16:08 UTC 2022 - José Iván López González <jlopez@suse.com>
+
+- Partitioner: PVs are not wrongly removed when resizing a VG
```
dgdavid, line 4: :thinking: Not sure if I got it. Not wrongly removed?

src/include/partitioning/ep-lvm-lib.rb
```diff
@@ -100,6 +100,53 @@ def EpCreateVolumeGroup
...
+
+      target_map.values.each do |device|
+        devices << device
+        devices += (device["partitions"] || [])
+      end
+
+      devices.find { |d| match_any_name?(d, device_name) }
+    end
+
+    # Checks wheher any of the device names matches with the given name
```
dgdavid:
```suggestion
    # Checks whether any of the device names matches with the given name
```

test/include/ep-lvm-lib_test.rb
```diff
@@ -0,0 +1,114 @@
...
+
+      it "returns the given name" do
+        expect(subject.device_kernel_name(target_map, device)).to eq(device)
+      end
+    end
+
+    context "when the given name is a unknown kernel name" do
+      let(:device) { "/dev/mapper/mpathnz" }
+
+      it "returns the given name" do
```
dgdavid, line 85: Is that true?

joseivanlopez: Yes. The idea is to return the kernel name if you find the device, otherwise the method will simply return the name you passed. It is commented in the description of the method, too.

package/yast2-storage.changes
joseivanlopez: Yes, one bad thing happening in the bug report is that some PVs are removed, but they shouldn't.

dgdavid: Ah! Sorry, I got it now. You mean that the change is not removing PVs by mistake. Thanks!

shundhammer (approved):
LGTM in general.

One NP point is that "kernel name" is a bit misleading here. Yes, technically they are kernel device names. But since this is about LVM, it's not anything like `/dev/sda2` etc. which may be unstable (even between reboots), it's actually device mapper names like `/dev/mapper/something` which are explicitly assigned, so they cannot be unstable in the same sense.

Maybe add a comment to that effect in case anybody reading the code wonders about the stability of those names? Because that was a frequent source of trouble in the past for _normal_ kernel device names like `/dev/sda2`.

joseivanlopez:
> One NP point is that "kernel name" is a bit misleading here. Yes, technically they are kernel device names. But since this is about LVM, it's not anything like `/dev/sda2` etc. which may be unstable (even between reboots), it's actually device mapper names like `/dev/mapper/something` which are explicitly assigned, so they cannot be unstable in the same sense.

Actually, it is the kernel name, because the search is done for PVs. In general, the PVs have no `/dev/mapper` names. In the scenario of the bug report, the PVs are `/dev/mapper/something` because they are multipath devices. Note that the PVs names are not stored in any file in the system, they are simply probed when analyzing the storage devices.

shundhammer:
OK then.

joseivanlopez:
https://build.suse.de/request/show/270875

